PR #7, opened 2020-05-30
This is the new version of m4e

feeds:


feeds:
# Todo

- [x] One issue is that the config file in django config is not resolved correctly `backend/match4everyone/config/participant_info.json`. Maybe do that with settings?

- [x] Translations are actually all there. Somehow also fails bc its trying to do sth with `00_old_m4h_matching_code`.

- [x] Travis einrichten

Baschdl:
Travis says `RuntimeError: Model class match4everyone-backend.apps.matching.models.user.User doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.`

feeds:
> Travis says `RuntimeError: Model class match4everyone-backend.apps.matching.models.user.User doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.

Could it be that in backend/Dockerfile 
the working directory is wrong?
It says `WORKDIR /match4everyone-backend`

feeds:
It actually wasnt ;)
It was the init in the backend... my pycharm tends to create this when I press "refactor" :sweat_smile:

kevihiiin (approved):
Looks good and the basic functionality works.
I would like to add functions from 'Cookiecutter Django'  (#11) or even let it generate some project skeleton but that can be done afterwards.

PR #14, opened 2020-06-06
Add new favicon

### LINKED ISSUE #4: Add a random logo
In general, the favicon is still from m4h, the random current logo is shitty

## Todo
- create a random logo
- create all favicon and logo svg's and pngs

## Resources
- `backend/static/img/favicon` and `backend/static/img/logo.svg`

bjrne:
Closes #4


maltezacharias (commented):
We should review whether 
https://www.npmjs.com/package/favicons-webpack-plugin

might be a better approach, as we will be using webpack for JavaScript, this could be used as well and generate all the favicons on deployment. Will test and report back

bjrne:
Sounds good, I was looking for something similar, but sadly there was nothing in python. Forgot to check js though :wink:

bjrne:
I think this would be out of scope for this PR. I have opened a new issue for that: #24 .

Let's merge this for now in order to have an icon :)

Baschdl:
@bjrne Do you still want to merge this or is it obsolete with the new logo or js?

bjrne:
Until we have a better one, in merge-ready state, yes.

backend/templates/base.html
```diff
@@ -15,23 +15,30 @@
...
-    <link rel="apple-touch-icon" sizes="57x57" href="{% static 'img/favicon/apple-icon-57x57.png' %}">
-    <link rel="apple-touch-icon" sizes="60x60" href="{% static 'img/favicon/apple-icon-60x60.png' %}">
-    <link rel="apple-touch-icon" sizes="72x72" href="{% static 'img/favicon/apple-icon-72x72.png' %}">
-    <link rel="apple-touch-icon" sizes="76x76" href="{% static 'img/favicon/apple-icon-76x76.png' %}">
-    <link rel="apple-touch-icon" sizes="114x114" href="{% static 'img/favicon/apple-icon-114x114.png' %}">
-    <link rel="apple-touch-icon" sizes="120x120" href="{% static 'img/favicon/apple-icon-120x120.png' %}">
-    <link rel="apple-touch-icon" sizes="144x144" href="{% static 'img/favicon/apple-icon-144x144.png' %}">
-    <link rel="apple-touch-icon" sizes="152x152" href="{% static 'img/favicon/apple-icon-152x152.png' %}">
-    <link rel="apple-touch-icon" sizes="180x180" href="{% static 'img/favicon/apple-icon-180x180.png' %}">
-    <link rel="icon" type="image/png" sizes="192x192" href="{% static 'img/favicon/android-icon-192x192.png' %}">
```
Baschdl, line 28: Android disappeared?

Baschdl (changes requested):
Works in combination with #58 but maybe a bit big (that's my whole screen) 😄 
image

backend/templates/base.html
bjrne: Honestly, it takes a full SAP-team to do this right. It is a workable solution for now, and if you trust [this website](https://realfavicongenerator.net/favicon_checker?protocol=https&site=match4healthcare.de), there is at least some icon that Android understands. There is no easy python tool for this job, doing it manually is tedious and not worth its time since this would only be a placeholder icon for instances of our projects. I hope there are some good libraries for js and web stuff that do this right, repeatably, consistent and complete. Malte is looking into this. Until then, this should work for most scenarios.

bjrne: And btw, as you can see on the website, these links don't even work in m4h. There is some serious cleanup to be done and these lines have no function at the moment anyway I think.

Baschdl: The website only finds the manifest and no icon. I think we used [https://favicon.io/favicon-converter/](https://favicon.io/favicon-converter/) before unless @andyz02 changed them in the meantime. [https://match4healthcare.de/static/img/favicon/android-icon-192x192.png](https://match4healthcare.de/static/img/favicon/android-icon-192x192.png) works for me.

maltezacharias: I believe that some of the files need to be available from the root dir, which won't work with the current statics setup :(
https://realfavicongenerator.net/ (mentioned by @Baschdl ) seems to be good. But I think we need to research this, whether it's possible (if even necessary) to make some of those files reachable from the app root. 

We _could_ always redirect or serve them with nginx but we **probably shouldn't** if at all possible.

bjrne: I think one way around the is to include lots of meta links in the html base header who link to the appropriate places, lots of favicon generators generate that html for you directly, so this should work. However, the current setup was and is flawed and this needs to be cleaned up nicely. I think a good moment for that would be when we switch to the automatic approach with a JS library that we envisioned @maltezacharias, right?

maltezacharias (approved):

backend/templates/base.html
maltezacharias: Yep. is there an issue for that? If not please create one. And for now we can merge this.

Baschdl: I set it up with a redirect in the very beginning of our project but I don't think we need it if we have the meta links. The 404 for `favicon.ico` shouldn't be a problem because the browser finds the appropriate icons with the meta links, see https://stackoverflow.com/a/59715513/3342058

Baschdl (dismissed):
The logo is still a bit too big

PR #18, opened 2020-06-14
Integrate Webpack for Javascript builds

### LINKED ISSUE #17: [JS] Implement webpack build process
**Topic:**
Implement webpack building process for javascript ressources

**Relevant issues:**

**Further description:**
Separates JS, CSS and used resources from the django backend. Provides the infrastructure for possible future integrations of modern javascript, e.g. Vue.js

**Alternative approaches that could be considered:**
Manually keep everything together and despair

**Todos**
- [X] Integrate into build process (travis)
- [x] Pre-commit hook for building/linting the JS
- [x] Clean-Up Changes

maltezacharias:
Fixes #17, builds on #21 

# General idea 
Adds new Build process for Javascript
This PR implements a new way to handle Javascript resources in the application.
Javascript modules are written inside /frontend/src. A bundle is created for every module found directly in the src folder.

Django is enhanced with the webpack-loader, that provides a render bundle tag, which automatically loads the created bundles.

## Production mode
During Docker build (docker-compose -f docker-compose.yml -f docker-compose.prod.yml up --build) an intermediate frontend container is created, webpack is run, bundles are created and copied over to the final backend container. This way the finished container already has the bundles inside the image and no frontend container is needed.

## Development mode
A frontend docker container is provided in the override file (used by docker-compose up --build) and continues to run a webpack instance, which watches for file-changes in the frontend/src folder. When changes are detected the JS bundles are rebuilt automatically


README.md
```diff
@@ -119,6 +119,54 @@ Student has an attribute for the user, user has an attribute for the student, ..
...
+Javascript sources are located in frontend/src folder
+
+For local development node needs to be installed. For pure python development alternatively Docker can be used to build the javascript bundles. At the moment only production builds are supported via docker.
+
+### Build via docker
+
+Run the following commands in the main directory to setup the docker image
+```
+cd frontend
+docker build -t frontend:latest .
```
Baschdl: Why don't you let `docker-compose` do the tagging? Why add another manual step at all?

README.md
```diff
@@ -119,6 +119,54 @@ Student has an attribute for the user, user has an attribute for the student, ..
...
+```
+
+Run `docker run -ti -v "$(pwd)/dist:/app/dist" --rm frontend:latest` to build the bundles. They will then be available under the dist directory. Docker-Compose files for the main app use volumes instead to share the files between containers.
+
+### Build (local install) dev & prod
+
+To build locally node needs to be installed. We suggest using [Node Version Manager](https://github.com/nvm-sh/nvm) to install node. Dependencies can be installed using `cd frontend && npm install`
```
Baschdl: I like to supply a working dev environment but most people won't use it, so it's debatable

backend/templates/impressum.html
```diff
@@ -29,8 +21,7 @@ <h1>Impressum</h1>
                     <h5>Personen hinter dem Projekt</h5>
                     <strong>match4everyone e.V.</strong><br><br>
                     Tel.: ---- <br>
-                    E-Mail:
-                    <script type="text/javascript">document.write(decode("abc@example.qr"))</script>
+                    E-Mail: "abc@example.qr"
```
Baschdl: Why do you remove the spam protection?

backend.dev.env.example
```diff
@@ -0,0 +1,2 @@
+# Token from mapbox.com to get Open Street Map tiles
+MAPBOX_TOKEN=
```
Baschdl, line 2: No `SECRET_KEY`?

Baschdl (changes requested):
This could have been more than one PR

README.md
bjrne: I would appreciate it as well :) Even when you're not using it most of the time, it seems invaluable to me for new forks.

maltezacharias:
> This could have been more than one PR

Strong disagree. This is a javascript refactor. With @bjrne's PEP8/Flake we decided to go the same route. I'd rather have this all together in one PR to make clear that it's just a refactor towards a proper build process. 

I took great efforts to make the commits in logical and understandable units, but I don't see an advantage  in splitting this PR

backend/templates/impressum.html
maltezacharias: If we feel we need that, we can do it properly later. We can make an issue. But this code is just plain awful and undocumented. We always have the old repo to bring it back.

README.md
maltezacharias: Hmmm my intention was to describe what you would need to do, to setup a local environment while also documenting the dockerized way. Currently both are possible:
- install node and build a bundle in your local dev environment, which is usually preferred. If you have node already it works quite well and handles different dependencies in multiple projects well
- If you don't want to have a local node setup you can use the provided Dockerfile. If you run it in dev mode it bind mounts the JS source and will rebuild automatically on every change. There are some features (hotreloading the code) which I don't plan to support in Docker as I don't want to spend that additional time.

I checked that both ways work, is the documentation unclear?

README.md
maltezacharias: This can be improved. I took care to copy the commands I used while setting it all up in the documentation because they were helpful for me and it felt quicker this way for example when using a local runserver plus frontend building docker. 

I think we should invest quite a bit of time in documenting the whole docker setup and our design consideration. I think the current state is mergeable but not finished. But I'd like to split future improvements in smaller PRs focusing on the pain areas

backend.dev.env.example
maltezacharias: The dev secret key is hardcoded. I created this file with the miminum necessary contents to successfully run a local development docker. This was broken with the introduction of MAPBOX_TOKEN.

maltezacharias:
This is quite a big refactoring change. I would like to merge this now as it is in a workable state and provides a good base for several other enhancements to build upon. I'd suggest to check if you can run it and if the documentation is sufficient and then merge it, so that we can better divide the next steps between us.

@Baschdl @kevihiiin @feeds what do you think about that approach?

README.md
bjrne: Hm, I have to say as someone who has barely worked with node before, I would be really unsure what to install here in which scenario. I don't really have a better suggestion, but this does not work for me. One solution could be to move the node stuff closer to the "normal" setup.

Possible open questions after reading this are:
- Do I need to install node when I want to develop a new version, but will only change django/python stuff in order to build the packs?
- What is the difference between js dev and production builds? (_At the moment only production builds are supported via docker._)
- Do I need to call special commands in order to compile JS?

maltezacharias: >  move the node stuff closer to the "normal" setup.
Meaning?

Baschdl:
> Strong disagree. This is a javascript refactor. With @bjrne's PEP8/Flake we decided to go the same route. I'd rather have this all together in one PR to make clear that it's just a refactor towards a proper build process.

A really strong disagree this PR has Docker changes, (html content+translation changes) and javascript changes which would have been perfectly separable

Baschdl:
> provides a good base for several other enhancements to build upon

One can always base a new branch on this branch but I'll give it a second look later on

bjrne:
I agree with @Baschdl, docker would have been a good PR on its own. The discussions about details just get a bit out of hand on one of this size. 

And btw, PEP8 was 3+ PRs ;)

EDIT: and for me personally, it's also more rewarding getting smaller PRs merged faster instead of big discussions opening on a PR that has been worked on for several weeks, which you just want to be done with.

README.md
Baschdl: You're right I also forgot to change the README in https://github.com/match4everyone/match4everything/pull/21. I'm ok-ish with merging this but would prefer a clean version from the start

backend/templates/impressum.html
Baschdl: I would not remove features because the code is undocumented. Do it the other way round; open an issue to improve this and leave it in like this

backend/apps/matching/templates/registration/logout.html
```diff
@@ -21,7 +21,7 @@
 	}
 
 </style>
-<script type="text/javascript">
+<script>
```
bjrne, line 24: Out of curiosity, why is this removed/what does it normally do?

maltezacharias: As per https://javascript.info/hello-world (probably one if not the best JS-resource I know, except for the MDN reference) both type and language attributes of the script tag are no longer required. So in essence it doesn't do anything. 

See also https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script :  
>The HTML5 specification urges authors to omit the attribute rather than provide a redundant MIME type.

There are still uses for the type attribute (using modules in Javascript, which is supported by most modern but not all of the used browsers, one of the reasons to use webpack) and there are several other interesting attributes to the tag. In our case none of them are needed.

backend/templates/impressum.html
maltezacharias: Revert 252692f8188ada8b4a63515c056d51fee2ec7b57

maltezacharias:
@Baschdl this now based on easier-deploy branch

maltezacharias:
Waiting for #21 to be merged

maltezacharias:
Will recreate this on a new branch for improved reviewability

PR #20, opened 2020-06-15
[Bugfix] Enable creation of superuser again

bjrne:
Also adds a new check:

```python
(is_A or is_B) == is_participant
```

And a testcase! :tada:


bjrne:
We could also use this chance to squash the migrations once again. What do you think?

feeds:
With regards to squashing - I would always regenerate and overwrite the migrations until we start supporting an active project, which will not happen in the next two months I think

feeds (approved):
Pretty, thanks for fixing this. I added a commit with squashed migrations :)

PR #21, opened 2020-06-15
Easier deployment

### LINKED ISSUE #57: backend errors (again) when database is starting the first time
Following error occurs, when deleting `database/data` and starting for the first time:
```
django.db.utils.OperationalError: could not connect to server: Connection refused
	Is the server running on host "database" (172.21.0.2) and accepting
	TCP/IP connections on port 5432?
``

### PR DESCRIPTION
**Pull request by Baschdl**
Problems:
- [X] `compilemessages` changes .mo file and triggers invalidation of docker cache despite of no source code changes

EDIT by malte:
- [X] Fix DB startup race condition causing error `FATAL:  the database system is starting up`
- [x] Address Bastis remarks about gosu
- [x] Add sample gateway error pages & nginx configuration (copy from [m4h PR 538](https://github.com/match4everyone/match4healthcare/pull/538))
- [x] Make docker container bind to localhost (see [m4h PR 540](https://github.com/match4everyone/match4healthcare/pull/540))
- [X] Fix backup script (implement match4everyone/match4healthcare@f0737b25f4f13cc6fef74bcb9517c1c87c91f35d)

Closes #57

### CODE REVIEW

File: backend/Dockerfile [hunk 1]
```diff
@@ -1,11 +1,18 @@
...
 RUN apt-get update && apt-get install -y python3 python3-pip libpq-dev gettext
+
 WORKDIR /match4everyone-backend
 COPY requirements.txt /match4everyone-backend/requirements.txt
 RUN pip3 install -r requirements.txt
 COPY requirements.prod.txt /match4everyone-backend/requirements.prod.txt
 RUN pip3 install -r requirements.prod.txt
 
+COPY . .
+RUN django-admin makemessages --no-location --ignore 00_old_m4h_matching_code
```
bjrne: Lets just remove the old code? I think the refactoring is mostly done now.

Baschdl: Separate PR or in this one?

Baschdl: https://github.com/match4everyone/match4everything/pull/26

Baschdl:
`Website reachable:	Website error code: 500Failed` not sure why. Only thing I see in the travis logs is that it doesn't print `worker started (pid: XYZ)`

maltezacharias (commented):
Looks good, I'll test tomorrow and approve after

maltezacharias:
Ahh, leave that error for me, I had something similar and set up a local travis like build env. I'll see what I can do about that

backend/Dockerfile
bjrne: #26 done, remember to remove this flag @maltezacharias

maltezacharias:
About the compilemessages problem... Is it possible for us to hook into that command and just remove the damn ChangedAt line? We have a commit history for that file anyway so that header field doesn't has a purpose anyway. I know we talked about this earlier but it is a constant nuisance everytime I test local deployment with docker

Baschdl:
> Is it possible for us to hook into that command and just remove the damn ChangedAt line?

`compilemessages` or better GNU `gettext` doesn't provide a way, it's discussed a lot online but I couldn't find a solution

Baschdl:
I mean we could implement something like https://github.com/sphinx-doc/sphinx/pull/3490 but that's maybe a bit overkill. We could just delete this line with `sed` in docker but this would also require that nobody commits this line again. @bjrne Would a similar (or the same) deletion before committing be possible with a pre-commit hook (would only work on unix @maltezacharias)?

bjrne:
It would be a very custom fix, but I can see that happening, yes. If I/we write that script for the hook for example in python, it would be cross-platform like all the other hooks :) If you want me to work on this, I can happily give this a try.

bjrne:
But I suspect even with the different approach, it won't get much shorter than in the sphinx link you posted...

But there would probably be a lot of people who would like to use this hook, given the volume of comments about this issue on the internet 😅

Baschdl:
The sphinx approach compares if there is any change, I would try if just deleting works for us but the comparison is definitively the cleaner version. @bjrne I'm happy if you implement it

maltezacharias:
Actually makemessages introduces the constant changes.
I have overridden the command to always set the same POT-Creation-Date. Now the file should only be detected as changed when there are actual changes to the content

maltezacharias:
> > Is it possible for us to hook into that command and just remove the damn ChangedAt line?
> 
> `compilemessages` or better GNU `gettext` doesn't provide a way, it's discussed a lot online but I couldn't find a solution

I made one. It's pretty easy actually, you can simply create a new command with the same name which will override the default one. I inherited my implementation from the original makemessages and modify the POT header. (POTs  are only temporarily created but the POT-Created Header is copied to the final .po file.

maltezacharias:
Okay, so I tried to get this to run and it was a lot more difficult than I expected. In the end there were a lot of changes and we can split the PR if you feel that would be more appropriate. 

I've made all my commits in a way that it should be easily understandable what I changed and why. With all those changes the following scenarios are supported:

- Local docker development using `docker-compose` without -f parameter using a bind mount to be able to change the code and runserver
- Production mode using `docker-compose -f docker-compose.yml -f docker-compose.prod.yml`, the rename was necessary as the bind mount for dev uses a docker override file and this way it is more easily run
- Production mode with a bind mount to the code is not yet supported, it could most easily be added with a third docker-compose file ( like for example `docker-compose -f docker-compose.yml -f docker-compose.prod.yml -f docker-compose.prod-bind.yml` 
- It is possible to set a user with a .env file that is used to run the containers. That whole part is quite difficult, as the container filesystem permissions pose a big problem. When we build the container the UID is not yet known (otherwise you would need a tailored image for every system built for the specific UID you're going to use) The solution for now is to set appropriate permissions on the user writeable directories. Unfortunately this can easily stop working if you start the container once as root accidentally and is rather difficult to repair afterwards. 
Using the .env file you can set the container to run with the UID you want before you start it the first time and it will automatically always start with the correct user.

scripts/write_envs_to_file.sh
```diff
@@ -1,10 +1,16 @@
...
-echo "POSTGRES_DB=${POSTGRES_DB}" > database.prod.env
-echo "POSTGRES_USER=${POSTGRES_USER}" >> database.prod.env
-echo "POSTGRES_PASSWORD=${POSTGRES_PASSWORD}" >> database.prod.env
-
-echo "SECRET_KEY=${SECRET_KEY}" > backend.prod.env
-echo "SENDGRID_API_KEY=${SENDGRID_API_KEY}" >> backend.prod.env
-echo "RABBITMQ_DEFAULT_USER=${RABBITMQ_DEFAULT_USER}" >> backend.prod.env
-echo "RABBITMQ_DEFAULT_USER=${RABBITMQ_DEFAULT_USER}" >> backend.prod.env
+DATABASE_ENV_VARS="POSTGRES_DB POSTGRES_USER POSTGRES_PASSWORD"
+BACKEND_ENV_VARS="SECRET_KEY SENDGRID_API_KEY SLACK_LOG_WEBHOOK LEAFLET_TILESERVER MAPBOX_TOKEN"
```
maltezacharias, line 4: I needed to add the new env vars and DRY happened

maltezacharias:
> But I suspect even with the different approach, it won't get much shorter than in the sphinx link you posted...
> 
> But there would probably be a lot of people who would like to use this hook, given the volume of comments about this issue on the internet 😅

@bjrne I think I've solved that easier than with pre-commit. See bc4943bb10b3837f41be38eb7d7e239c193389e8

maltezacharias:
I just read another blog post: https://denibertovic.com/posts/handling-permissions-with-docker-volumes/ this is actually not a bad approach. We could run the container as root, make sure that our permissions are set up correctly, if needed create passwd and groups entries in the entrypoint, but then drop privileges with gosu

backend/entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+  done
+fi
+
+# While collectstatic should rather run in the Dockerfile build
+# it depends on the right DJANGO_SETTINGS_MODULE as it is subject to the
+# STATICFILES_STORAGE setting. Whitenoise is only configured in production as of now
+# so we would need to build the image in production mode, this would require setting a DJANGO_SECRET key
+echo PYTHONPATH: $PYTHONPATH
+django-admin collectstatic --no-input
+django-admin migrate
```
kevihiiin, line 16: Do we want to run migrate every time the container is started?
I know that in our use-case we migrate every time we restart the container because the build changes. But in general it feels wrong to touch the database structure everytime the container is just started even if we don't have to.
A problematic scenario: Say after an unexpected shutdown the database gets corrupted and we now try to re-migrate on startup.

docker-compose.prod.yml
```diff
@@ -4,7 +4,7 @@ services:
 
   backend:
     volumes:
-      - './backend:/match4everyone-backend'
+      - backend-run:/backend/run
```
maltezacharias: We should probably create a subdirectory for logs and only bind that to a volume
and exclude the statics.

backend/Dockerfile
```diff
@@ -1,11 +1,23 @@
...
 
+COPY . .
+
+RUN python3 manage.py makemessages --no-location --ignore 00_old_m4h_matching_code
+RUN python3 manage.py compilemessages
+
+# Change permissions on run/ in case app is later run by non-root user
+# and delete log files as these are created during the above commands when django loads the configuration
+# and will be created non writeable by others than root
+RUN chmod -Rc a+rwX run && rm -v run/*log
```
maltezacharias: Better solution gosu or su-exec, this should be changed to:
1. ENV variable with uid and gid to be used by container
2. run container as root (remove user param from yml)
3. create user and group inside of container with fixed names (e.g. backend)
4. execute the processes using gosu/su-exec with user from 3

backup.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 source database.prod.env
```
maltezacharias: needs testing!

README.md
```diff
@@ -7,51 +7,47 @@ Originally developed as [match4healthcare](https://github.com/match4everyone/mat
...
+After changes to the Docker configuration, you have to restart and build the containers with `docker-compose up --build`.
 
 ### Production
-Set `SECRET_KEY`, `SENDGRID_API_KEY` and `MAPBOX_TOKEN`in `backend.prod.env` for Django
-`POSTGRES_DB`, `POSTGRES_USER`, `POSTGRES_PASSWORD`  inside `database.prod.env` for postgres on your host machine.
-Also add a `SLACK_LOG_WEBHOOK` to enable slack logging.
+Copy `backend.prod.env.example` to `backend.prod.env` and set variables as documented in the example file for Django
+Copy `database.prod.env.example` to `database.prod.env` and set variables as documented in the example file for postgres on your host machine.
 
 To run a container in production and in a new environment execute the `deploy.sh` script which builds the containers, runs all configurations and starts the web service.
```
Baschdl:
```suggestion
To run a container in production and in a new environment execute the `docker-compose -f docker-compose.yml -f docker-compose.prod.yml up -d --build` script which builds the containers, runs all configurations and starts the web service.
```

Baschdl:
> We could run the container as root, make sure that our permissions are set up correctly, if needed create passwd and groups entries in the entrypoint, but then drop privileges with gosu

@maltezacharias Not sure how comfortable I am with using `gosu` without understanding it 100% (See https://github.com/tianon/gosu/issues/37 and https://github.com/ncopa/su-exec/issues/23). Running with `exec gosu ...` should be fine but still...

Baschdl:
I get a `django.db.utils.OperationalError: FATAL:  the database system is starting up`-error when starting with `docker-compose -f docker-compose.yml -f docker-compose.prod.yml up -d --build` and not having a database set up previously. After waiting for the db and executing `docker-compose -f docker-compose.yml -f docker-compose.prod.yml up -d --build`  again, I get a `django.db.utils.OperationalError: FATAL:  no pg_hba.conf entry for host [host], user [user], database [db], SSL off`-error.

This only happens sometimes, so we seem to have a race condition somewhere.

Baschdl:
On a positive note, #58 isn't a problem here

kevihiiin:
EDIT by Malte: Removed Todos and added at top so we don't miss them

maltezacharias:
> I get a `django.db.utils.OperationalError: FATAL: the database system is starting up`-error when starting with `docker-compose -f docker-compose.yml -f docker-compose.prod.yml up -d --build` and not having a database set up previously. After waiting for the db and executing `docker-compose -f docker-compose.yml -f docker-compose.prod.yml up -d --build` again, I get a `django.db.utils.OperationalError: FATAL: no pg_hba.conf entry for host [host], user [user], database [db], SSL off`-error.
> 
> This only happens sometimes, so we seem to have a race condition somewhere.

I think I read somewhere in the postgres container docs that this is to be expected on the first startup when I researched the user problem. Thanks for the head-up. What kind of specs is the system you're running that on? I somehow doubt that my setup is that much faster.

Found it:
> **Caveats**
> If there is no database when postgres starts in a container, then postgres will create the default database for you. While this is the expected behavior of postgres, this means that it will not accept incoming connections during that time. This may cause issues when using automation tools, such as docker-compose, that start several containers simultaneously.

I think there is no way to prevent it, at least I can't think of one. And I'm not sure if this is actually the reason this fails, because the way this sounds the DB should not be reachable at that time and in the new entrypoint script we make sure it is reachable before we do anything else in the backend container. This puzzles me...

**EDIT**: I was able to recreate the error by modyfying my script and removing the connection check. It is still not possible to get that error everytime for me. I think I figured it out though, our entrypoint script checks if the DB port is reachable. Postgres seems to start to listen to that port during DB startup even if it's not yet ready to process requests. When you connect to Postgres during this short window you get that error message. 

**EDIT2**: After doing some more research I think I found the culprit:
https://www.postgresql.org/docs/current/hot-standby.html
> When the hot_standby parameter is set to true on a standby server, it will begin accepting connections once the recovery has brought the system to a consistent state. All such connections are strictly read-only; not even temporary tables may be written.

That is probably what's happening. The server is coming up in hot_standby mode and migrate tries to modify the data causing the error message. I've captured the log from my attempt:

```
➜  match4everything git:(easier-deploy) ✗ docker-compose -f docker-compose.yml -f docker-compose.prod.yml down && docker-compose -f docker-compose.yml -f docker-compose.prod.yml up
Removing match4everything_backend_1  ... done
Removing match4everything_database_1 ... done
Removing network match4everything_network
Creating network "match4everything_network" with the default driver
Creating match4everything_database_1 ... done
Creating match4everything_backend_1  ... done
Attaching to match4everything_database_1, match4everything_backend_1
backend_1   | PYTHONPATH: /backend
database_1  | The files belonging to this database system will be owned by user "postgres".
database_1  | This user must also own the server process.
database_1  | 
database_1  | The database cluster will be initialized with locale "en_US.utf8".
database_1  | The default database encoding has accordingly been set to "UTF8".
database_1  | The default text search configuration will be set to "english".
database_1  | 
database_1  | Data page checksums are disabled.
database_1  | 
database_1  | fixing permissions on existing directory /var/lib/postgresql/data ... ok
database_1  | creating subdirectories ... ok
database_1  | selecting dynamic shared memory implementation ... posix
database_1  | selecting default max_connections ... 100
database_1  | selecting default shared_buffers ... 128MB
database_1  | selecting default time zone ... Etc/UTC
database_1  | creating configuration files ... ok
database_1  | running bootstrap script ... ok
database_1  | performing post-bootstrap initialization ... ok
database_1  | syncing data to disk ... initdb: warning: enabling "trust" authentication for local connections
database_1  | You can change this by editing pg_hba.conf or using the option -A, or
database_1  | --auth-local and --auth-host, the next time you run initdb.
database_1  | ok
database_1  | 
database_1  | 
database_1  | Success. You can now start the database server using:
database_1  | 
database_1  |     pg_ctl -D /var/lib/postgresql/data -l logfile start
database_1  | 
database_1  | waiting for server to start....2020-07-01 23:59:47.725 UTC [46] LOG:  starting PostgreSQL 12.3 (Debian 12.3-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit
database_1  | 2020-07-01 23:59:47.728 UTC [46] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
database_1  | 2020-07-01 23:59:47.785 UTC [47] LOG:  database system was shut down at 2020-07-01 23:59:47 UTC
database_1  | 2020-07-01 23:59:47.791 UTC [46] LOG:  database system is ready to accept connections
database_1  |  done
database_1  | server started
database_1  | CREATE DATABASE
database_1  | 
database_1  | 
database_1  | /usr/local/bin/docker-entrypoint.sh: sourcing /docker-entrypoint-initdb.d/init-db.sh
database_1  | ALTER ROLE
database_1  | ALTER ROLE
database_1  | ALTER ROLE
database_1  | 
database_1  | 2020-07-01 23:59:48.030 UTC [46] LOG:  received fast shutdown request
database_1  | waiting for server to shut down....2020-07-01 23:59:48.033 UTC [46] LOG:  aborting any active transactions
database_1  | 2020-07-01 23:59:48.034 UTC [46] LOG:  background worker "logical replication launcher" (PID 53) exited with exit code 1
database_1  | 2020-07-01 23:59:48.035 UTC [48] LOG:  shutting down
database_1  | 2020-07-01 23:59:48.060 UTC [46] LOG:  database system is shut down
database_1  |  done
database_1  | server stopped
database_1  | 
database_1  | PostgreSQL init process complete; ready for start up.
database_1  | 
database_1  | 2020-07-01 23:59:48.149 UTC [1] LOG:  starting PostgreSQL 12.3 (Debian 12.3-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit
database_1  | 2020-07-01 23:59:48.149 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
database_1  | 2020-07-01 23:59:48.149 UTC [1] LOG:  listening on IPv6 address "::", port 5432
database_1  | 2020-07-01 23:59:48.167 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
database_1  | 2020-07-01 23:59:48.220 UTC [73] LOG:  database system was shut down at 2020-07-01 23:59:48 UTC
database_1  | 2020-07-01 23:59:48.220 UTC [74] FATAL:  the database system is starting up
backend_1   | Traceback (most recent call last):
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/db/backends/base/base.py", line 220, in ensure_connection
backend_1   |     self.connect()
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/utils/asyncio.py", line 26, in inner
backend_1   |     return func(*args, **kwargs)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/db/backends/base/base.py", line 197, in connect
backend_1   |     self.connection = self.get_new_connection(conn_params)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/utils/asyncio.py", line 26, in inner
backend_1   |     return func(*args, **kwargs)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/db/backends/postgresql/base.py", line 185, in get_new_connection
backend_1   |     connection = Database.connect(**conn_params)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/psycopg2/__init__.py", line 127, in connect
backend_1   |     conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
backend_1   | psycopg2.OperationalError: FATAL:  the database system is starting up
backend_1   | 
backend_1   | 
backend_1   | The above exception was the direct cause of the following exception:
backend_1   | 
backend_1   | Traceback (most recent call last):
backend_1   |   File "/usr/local/bin/django-admin", line 11, in <module>
backend_1   |     sys.exit(execute_from_command_line())
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
backend_1   |     utility.execute()
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/core/management/__init__.py", line 395, in execute
backend_1   |     self.fetch_command(subcommand).run_from_argv(self.argv)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/core/management/base.py", line 328, in run_from_argv
backend_1   |     self.execute(*args, **cmd_options)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/core/management/base.py", line 369, in execute
backend_1   |     output = self.handle(*args, **options)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/core/management/base.py", line 83, in wrapped
backend_1   |     res = handle_func(*args, **kwargs)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/core/management/commands/migrate.py", line 86, in handle
backend_1   |     executor = MigrationExecutor(connection, self.migration_progress_callback)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/db/migrations/executor.py", line 18, in __init__
backend_1   |     self.loader = MigrationLoader(self.connection)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/db/migrations/loader.py", line 49, in __init__
backend_1   |     self.build_graph()
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/db/migrations/loader.py", line 212, in build_graph
backend_1   |     self.applied_migrations = recorder.applied_migrations()
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/db/migrations/recorder.py", line 76, in applied_migrations
backend_1   |     if self.has_table():
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/db/migrations/recorder.py", line 56, in has_table
backend_1   |     return self.Migration._meta.db_table in self.connection.introspection.table_names(self.connection.cursor())
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/utils/asyncio.py", line 26, in inner
backend_1   |     return func(*args, **kwargs)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/db/backends/base/base.py", line 260, in cursor
backend_1   |     return self._cursor()
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/db/backends/base/base.py", line 236, in _cursor
backend_1   |     self.ensure_connection()
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/utils/asyncio.py", line 26, in inner
backend_1   |     return func(*args, **kwargs)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/db/backends/base/base.py", line 220, in ensure_connection
backend_1   |     self.connect()
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/db/utils.py", line 90, in __exit__
backend_1   |     raise dj_exc_value.with_traceback(traceback) from exc_value
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/db/backends/base/base.py", line 220, in ensure_connection
backend_1   |     self.connect()
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/utils/asyncio.py", line 26, in inner
backend_1   |     return func(*args, **kwargs)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/db/backends/base/base.py", line 197, in connect
backend_1   |     self.connection = self.get_new_connection(conn_params)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/utils/asyncio.py", line 26, in inner
backend_1   |     return func(*args, **kwargs)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/django/db/backends/postgresql/base.py", line 185, in get_new_connection
backend_1   |     connection = Database.connect(**conn_params)
backend_1   |   File "/usr/local/lib/python3.6/dist-packages/psycopg2/__init__.py", line 127, in connect
backend_1   |     conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
backend_1   | django.db.utils.OperationalError: FATAL:  the database system is starting up
backend_1   | 
database_1  | 2020-07-01 23:59:48.228 UTC [1] LOG:  database system is ready to accept connections
match4everything_backend_1 exited with code 1
^CGracefully stopping... (press Ctrl+C again to force)
Stopping match4everything_database_1 ... done
```

maltezacharias:
> @maltezacharias Not sure how comfortable I am with using `gosu` without understanding it 100% (See [tianon/gosu#37](https://github.com/tianon/gosu/issues/37) and [ncopa/su-exec#23](https://github.com/ncopa/su-exec/issues/23)). Running with `exec gosu ...` should be fine but still...

I read up on that. It is a tricky enough security issue that I think I understand it but it's rooted to deply in the ioctls it's based on that I can fully understand it or even produce an example exploit. The discussions I read make me believe it should be safe for our use. Alternatively we can try to use su. I think the main difference is that we would then have the su process running. I haven't found a reference to any other disadvantages so far. Will test

maltezacharias:
@kevihiiin please have a go and test it thoroughly. It works nicely here, but @Baschdl found some strange bugs the last time I thought that.

maltezacharias:
I consider this PR complete and ready for review+merge. Bastis points about gosu are valid and merit further consideration. I suggest tackling that in a new PR as this PR already got a lot more complicated than we thought.

I am convinced though that we have resolved some hard to track down possible problems here and will also backport to m4h

maltezacharias:
https://github.com/match4everyone/match4everything/commit/c458215845d9ef1044df59e70c68ec764c07bab8 closes #57

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+      context: ./backend
+      dockerfile: Dockerfile
+    ports:
+      - '127.0.0.1:8000:8000'
+    tty: true
+    environment:
+      - DJANGO_SETTINGS_MODULE=match4everyone.settings.development
+      - CI
+      - TRAVIS
+      - TRAVIS_PULL_REQUEST_SLUG
```
kevihiiin: Do we need to explicitly define these variables?
Docker passes them automatically when they are present on the host environment (travis).

When we are deploying not on travis, the variables are still specified (but empty)
```suggestion
```

backend/run/.gitignore
```diff
@@ -7,11 +7,16 @@
...
 # Except for the media folder
 !media
-!media/.gitkeep
 # But its content
 media/*
+!media/.gitkeep
+
+!log
+log/*
+!log/.gitkeep
```
kevihiiin: Git tells me it wants a newline at the end of this file
```suggestion
!log/.gitkeep

```

backend/Dockerfile
```diff
@@ -1,11 +1,28 @@
...
 RUN pip3 install -r requirements.prod.txt
 
+COPY . .
+
+# Set Django Settings to production, for the built docker image this will use whitenoise to collect the static files
+# Our app will refuse to run without a SECRET_KEY. collectstatic parses the configuration module and fails
+# if no key is set. To prevent this a dummy key is used. In production mode this needs to be supplied
+# using an env file in the docker-compose.yml
+ENV DJANGO_SETTINGS_MODULE="match4everyone.settings.production" SECRET_KEY="CzXG4ItUiwLUfTH2abQQ0qTzMSRiiDni"
+RUN python3 manage.py makemessages --no-location --ignore 00_old_m4h_matching_code
```
kevihiiin: In our readme we tell the user to use this command:
`[...] makemessages -l de --no-location`

Shouldn't we use the same in the docker file itself?

```suggestion
RUN python3 manage.py makemessages --no-location
```

backend/Dockerfile
```diff
@@ -1,11 +1,28 @@
...
+COPY requirements.prod.txt /backend/requirements.prod.txt
 RUN pip3 install -r requirements.prod.txt
 
+COPY . .
+
+# Set Django Settings to production, for the built docker image this will use whitenoise to collect the static files
+# Our app will refuse to run without a SECRET_KEY. collectstatic parses the configuration module and fails
+# if no key is set. To prevent this a dummy key is used. In production mode this needs to be supplied
+# using an env file in the docker-compose.yml
+ENV DJANGO_SETTINGS_MODULE="match4everyone.settings.production" SECRET_KEY="CzXG4ItUiwLUfTH2abQQ0qTzMSRiiDni"
```
kevihiiin: Given that this container is setup to run in development mode (the `runserver` command) we probably should also use settings.development here. 

I know it does not make any differences in our docker-compose setup as we override this but just to keep consistency.

Alternatively (which is what I would prefer) we setup this whole docker file to run as production and use gunicorn instead of "runserver". If we push it via DockerHub and someone uses it in standalone mode, it would be production ready by default.

.env.example
```diff
@@ -0,0 +1,3 @@
+# Use backend_user to specify a user to run the backend container as.
+# If none is specified root will be used
+BACKEND_USER=user:group
```
kevihiiin, line 3: Is the "human readable" username okay "dummy:dummy" or does it have to be an UID "42:42"?

README.md
```diff
@@ -7,51 +7,56 @@ Originally developed as [match4healthcare](https://github.com/match4everyone/mat
...
-`docker exec --env PYTHONPATH="/match4everyone-backend:$PYTHONPATH" backend django-admin compilemessages`
-4. Collect static
-`docker exec backend python3 manage.py collectstatic --no-input`
-5. Migrate
-`docker exec backend python3 manage.py migrate`
-5. Check if all the variables are correct
-`docker exec backend python3 manage.py check`
-6. Restart the backend container (important, whitenoise does not reload static files after it has started)
-`docker-compose -f docker-compose.dev.yml -f docker-compose.prod.yml down && docker-compose -f docker-compose.dev.yml -f docker-compose.prod.yml up -d`
+#### Build the containers
```
kevihiiin, line 47: This section does not really describe how to build it manually as the docker image runs the steps automatically. Maybe we should replace the `docker-compose` command with something without docker and rename the section to "Setup without docker" or something

backup.sh
```diff
@@ -1,4 +1,13 @@
...
+set -o errexit
+
+echo "Creating PostgreSQL Dump"
+docker-compose -f docker-compose.yml -f docker-compose.prod.yml exec database sh -c 'pg_dumpall -U $POSTGRES_USER> /backups/pg_backup-$(date +%F_%H%M%S).sql'
+DB_CONTAINER="$(docker-compose -f docker-compose.yml -f docker-compose.prod.yml ps -q database)"
+docker run --rm --volumes-from $DB_CONTAINER -v $(pwd):/backup-bind-mount alpine sh -c "mv -v backups/* /backup-bind-mount/database/backups"
+
+
+#Should we keep this? Have we ever verified this works?
+#docker-compose -f docker-compose.yml -f docker-compose.prod.yml exec backend  sh -c 'python3 manage.py dumpdata > /backend/backups/fixture-$(date +%F_%H%M%S).json'
```
kevihiiin: Given that we can load fixtures into Django with this method and this is the official way to dump it I think it works

backup.sh
```diff
@@ -1,4 +1,13 @@
...
-docker-compose -f docker-compose.dev.yml -f docker-compose.prod.yml exec backend sh -c 'python3 manage.py dumpdata > /match4everyone-backend/backups/fixture-$(date +%F_%H%M%S).json'
-docker-compose -f docker-compose.dev.yml -f docker-compose.prod.yml exec database sh -c "pg_dumpall -U $POSTGRES_USER> /backups/pg_backup-$(date +%F_%H%M%S).sql"
+# This backup script will create a database dump from the postgres container
+# the created backup will then be moved to the current directory
+set -o errexit
+
+echo "Creating PostgreSQL Dump"
+docker-compose -f docker-compose.yml -f docker-compose.prod.yml exec database sh -c 'pg_dumpall -U $POSTGRES_USER> /backups/pg_backup-$(date +%F_%H%M%S).sql'
+DB_CONTAINER="$(docker-compose -f docker-compose.yml -f docker-compose.prod.yml ps -q database)"
+docker run --rm --volumes-from $DB_CONTAINER -v $(pwd):/backup-bind-mount alpine sh -c "mv -v backups/* /backup-bind-mount/database/backups"
```
kevihiiin: What does this line do?:) Where is /backup-bind-mount actually mounted?
- - - 
Edit:
(docker run --rm --volumes [...])

kevihiiin (changes requested):

docker-compose.yml
kevihiiin: Needs to moved to .prod

kevihiiin (approved):

README.md
kevihiiin: @maltezacharias let's just leave it for now. The added information is good and the restructuring of the README should be part of #33 as it is a general issue.

backend/Dockerfile
maltezacharias: We can't use dev here, because then the upload post processing with whitenoise will not run.
Will evaluate switching to whitenoise in dev, I think that should be ok as well, because runserver won't require collectstatic but need to verify.

backup.sh
maltezacharias: /backup-bind-mount is inside the throw-away container. 
Basically this will create another container, mv the backups to the current wd and then exit and remove the container.
Will add comment in the backup.sh for clarification.

backup.sh
Baschdl: @maltezacharias I also don't see a disadvantage of using two different backup methods

README.md
```diff
@@ -7,51 +7,56 @@ Originally developed as [match4healthcare](https://github.com/match4everyone/mat
 
 ## Quick install
 - Copy `backend.prod.env.example` and `database.prod.env.example` to `backend.prod.env` and `database.prod.env` and fill in appropriate values
-- Run `./deploy.sh` (uses Docker) and visit `http://localhost:8000/`
+- Run `docker-compose -f docker-compose.yml -f docker-compose.prod.yml up -d --build` (uses Docker) and visit `http://localhost:8000/`
 
+## Environment variables
+
+TODO
```
Baschdl, line 14: A miracle that @bjrne's pre-commit didn't spot this 😝 @maltezacharias Did you intend to document it in this PR?

backend/Dockerfile
Baschdl: @maltezacharias We should also add an explanation to the readme

README.md
maltezacharias: Nope, I just noticed we never updated that section and we have an outstanding issue for the Readme/Doc anyway. I felt lucky when it didn't catch me :D. I did add a reference to the sample .envs instead, and checked the documentation in there

backup.sh
maltezacharias: Yep, there is none, I was just lazy. Will make that one work as well :)

Baschdl:
@maltezacharias can be merged after addressing the comments I also commented

maltezacharias:
Thx for the review

Baschdl:
We have to also create a `backend.dev.env` for travis

backend/Dockerfile
maltezacharias: I moved whitenoise to common configuration. It is thus no longer necessary to set this here and further documentation is not required.

Issue #22, opened 2020-06-15
Clarify meaning of "staff"


bjrne:
Currently, we have `Staff`-users and users with the built-in `is_staff`-flag set.

The `is_staff` flag determines, whether a user can log into the admin panel.

Our Staff user currently only has the boolean flag whether they can approve type a and b users. These will become obsolete with the Introduction of #12 , the permission management.

My proposal would be:
Remove the staff user once #12 is closed, and limit the creation of staff users to the admin panel, where you can easily tick the `is_staff` flag, and django even has a built-in permission for that.

bjrne:
The path forward is to remove our custom staff user and give the `is_staff`-flag to users who need administrative privileges. This still allows to have well-designed UI pages where they can approve users for example. The authorization can be implemented with fine-grained permissions if necessary.

feeds:
Most of the discussoin is now in #51

PR #23, opened 2020-06-15
Create permission system

bjrne:
Depends on #20 

Use djangos built-in permissions.
Using groups as permission holders and thinking of them as tags for this purpose is the way to use permissions proposed on many discussion sites.

Restricting a view to users with a certain permission works like this:
https://github.com/match4everyone/match4everything/blob/c5878424c542ee6ff7c11646ce35adaad5b69d3a/backend/apps/use_statistics/views.py#L26

If the groups and permissions should be created automatically, there are two main options:
1. via migration
2. via fixture

I chose the migration as it is more explicit in how things change and also does not require a conscious effort of the end user to set up. 

There is one slighty annoying thing with django permissions: they need a contenttype tag. Since django auto-generates crud-permissions for all models, I added the `approve`-permissions to their respective contenttype and used the `Staff`-contenttype for the rest. More options on how to handle this are presented [here](https://stackoverflow.com/a/13933002).


backend/apps/matching/models/user.py
```diff
@@ -49,11 +49,16 @@ class Meta(AbstractUser.Meta):
             # we can only have either participant a or participant b for one user
             models.CheckConstraint(
                 name="%(app_label)s_%(class)s_has_only_one_participant",
-                check=(~models.Q(is_A=False, is_B=False,)),
+                check=(~(models.Q(is_A=True) & models.Q(is_B=True))),
             ),
-            # only either staff or participant
+            # only either staff or participant or none
```
Baschdl: Is `none` sensible?

bjrne: That is a design choice - I just wanted to reflect what the test is currently testing for clarification. Design by @feeds , but I can imagine that these validations will still change over the upcoming weeks so I wouldn't think too much about it now. We should probably have another design-choices call soon, there are some more questions pending.

feeds (commented):
I think the permissions in the migration are very neat and easy to grasp. They would also be configurable upon creation of the system. 

I just worry that this will mix up with regenerating the migrations  (to only the __inital__ migration file) as the model evolves during our development. Do you have any suggestions on this?

bjrne:
I would suggest removing this migration for the sake of squashing, squash to one initial migration and then add this one back manually as the second migration, since schema and data migrations should be separate and would allow forks to remove this migration and automatically start from scratch. All these action could then in theory also be applied manually from the admin panel in django.

feeds:
But I think it is not very sustainable for our development process to have this "do-not-delete-migration" that does not look too different from the other ones if we change the model.

Also, someone that sets up the repository from scratch with their own config would have to initialize the migrations themselves - have you tried if everything works fine if there is no "initial" migration but the 002_... migration that sets up the permissions?

bjrne:
I think until we have the config, it would be premature optimization to think about the consequences, and even then I think this would still make sense. From a development perspective, unless we expect forks at this very moment, it also does not really make sense to squash migrations in every PR - it forces db-deletes unnecessarily and we could instead squash them prior to "releasing", so no further thinking is necessary.

Regarding people who set up this repository themselves: I think either they have a grasp on django and know what they are doing, or they don't. In either way, nobody would think of deleting file 0001 and leaving 0002 in place i guess. Everything in the migration can be done and undone manually in the admin panel. Adding more migrations for renaming purposes would also work, so I am not sure whether these cases really exist for us.

bjrne:
Removing the decorator will be part of a separate PR, since this touches the question of using class-based views and might need a discussion on its own. this is now ready to merge :+1: @feeds @josauder @Baschdl

backend/apps/matching/tests.py
```diff
@@ -1,14 +1,56 @@
...
+
+class MigrationTestCase(TransactionTestCase):
+    def test_migration_with_data(self):
+        # Migrate to starting point
         output = StringIO()
-        args = ["--noinput", "--email=staff@example.com", "--username=staff"]
-        opts = {}
-        call_command("createsuperuser", *args, stdout=output, **opts)
-        self.assertIn("Superuser created successfully", output.getvalue())
+        args = ["matching", "0001_initial"]
```
feeds, line 36: Could we maybe create a code test if the permission migration is still present? That way we would prevent accidental deletions. But no must have.

feeds (changes requested):
When I apply the migration I get a `django.db.utils.DatabaseError: malformed database schema (matching_user) - no such column: new__matching_user.is_A`. Weird that this does not happen in the travis test (is it a postgres thing?). But this stems from an older issue, the db constraints. When I delete the `matching_user_only_participant_if_part_of_a_or_b` I can generate the database just fine.

Otherwise everything looks good :)

backend/apps/matching/migrations/0002_permission_group_creation.py
```diff
@@ -0,0 +1,143 @@
+# Generated by Django 3.0.7 on 2020-06-15 12:28
+
+from django.db import migrations
+
+
+# This is added here and not in the project because these names are know to be valid only
```
maltezacharias, line 6: What is meant by "not in the project"? I am unfamiliar with the migrations. Can we go over them?

backend/apps/matching/migrations/0002_permission_group_creation.py
```diff
@@ -0,0 +1,143 @@
...
+# in at this very moment in time - later migrations could alter them.
+# Thus we only need them here for now. Saving these as constants elsewhere for easy usability
+# in the project is a different thing, but probably unnecessary when we can use decorators.
+class NewPermissions:
+    can_approve_type_a = "can_approve_type_a"
```
maltezacharias, line 11: Are we really committed to hardcode everything as A and B everywhere in the code. Is that really the most elegant way?

backend/apps/matching/models/staff.py
```diff
@@ -7,8 +7,6 @@ class Staff(models.Model):
     """An admin class for organizing the different access classes for maintainers of the platform."""
```
maltezacharias, line 7: Is Staff class necessary at all anymore? I don't see what it could be used for. 
I'd suggest removing the Model

backend/apps/use_statistics/views.py
```diff
@@ -23,6 +23,7 @@ def get_ttl_hash(seconds=300):
 
 @login_required
 @staff_member_required
+@permission_required("matching.can_view_access_stats")
```
maltezacharias, line 26: This is exactly what I had in mind. Beautiful!

maltezacharias (changes requested):
I think we should add some documentation of the roles and permissions in the README.md and a link to the relevant Django doc. I know this will blow up the README, but I'd rather have it in there now and move towards github pages in the next weeks.

I would remove the Staff model, I don't see what it would be necessary for. Is there a usecase? As far as I understand the is_staff property belongs to the User and will stay anyway?

backend/apps/matching/models/staff.py
bjrne: See #22, not scope of this issue

backend/apps/matching/migrations/0002_permission_group_creation.py
bjrne: Like the previous comment, this is something we should have another design decisions call again soon, I think for now it is the most appropriate solution until someone drafts something better, so I stuck with it.

bjrne:
I disagree with putting information about this into the readme.
1. Until now, there is barely any information on specific code decisions and designs in there, mostly setup. (And it is already bloated)
2. Information on _why_ we do (uncommon) things should we directly in the code, not the Readme. Information on _how_ we got there belongs to the issue or PR. 
3. We use the standard, best practise django way, the links are already linked in the issue and in the migration and among the first google search results.
4. When this gets adopted by other parts of the code we could think about how to extend the documentation further, but the permissions and roles might still change quite a bit, so the Readme feels like the wrong place right now.

backend/apps/matching/migrations/0002_permission_group_creation.py
bjrne: Let's discuss it in the next call

backend/apps/matching/tests.py
bjrne: Nice idea. Just a `file-exists`-test? But actually the only improvement I can see would be the more specific error message, since the test you linked test would automatically fail too :)

maltezacharias:
I still think we should document the basis permissions concept that we use, if not in Readme then in Contributing / Coding guidelines. This should and will form basis of the coding guidelines for this project and it doesn't make sense put this neither in the source code because it will be used in many different parts nor only in the issues because for basic dev guidelines you shouldn't need to scan all the issues of a project. For now we should collect these little bits. Unless somebody volunteers to go over all issues and PRs that will be created in the next few months to collect documentation. We discussed in the call that the docs might be exceedingly helpful to whoever will use this later.

feeds:
> When I apply the migration I get a `django.db.utils.DatabaseError: malformed database schema (matching_user) - no such column: new__matching_user.is_A`. Weird that this does not happen in the travis test (is it a postgres thing?). But this stems from an older issue, the db constraints. When I delete the `matching_user_only_participant_if_part_of_a_or_b` I can generate the database just fine.
> 
> Otherwise everything looks good :)

@bjrne what is your take on this?

Baschdl:
I'm with @maltezacharias regarding documenting in a setup instruction at least what the different permissions mean as this is something every person has to decide for their users

feeds (approved):
Okay, its an issue my system..

maltezacharias (approved):

PR #26, opened 2020-06-15
Delete code from match4healthcare #git

Baschdl:


bjrne (approved):
but I wanted that -37,988 diff on my stats :cry: 

:grin:

Issue #9, opened 2020-05-30
User signup results in django.db.utils.DataError


Baschdl:
Signing up as User A results in `django.db.utils.DataError: value too long for type character varying(3)`

kevihiiin:
The error does only occur when using the Docker container.
I tried to run the system without Docker but with production settings (and SQLite instead of postgres) and it works.

So either docker or postgres is the issue

Baschdl:
My best guess would be postgres from the error message

Baschdl:
Doesn't happen anymore

PR #27, opened 2020-06-15
Use user 'postgres' for postgres container

Baschdl:
Seems to work but currently not really testable with https://github.com/match4everyone/match4everything/issues/9

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -32,6 +32,7 @@ services:
      - database.prod.env
     expose:
      -  5432
+    user: postgres
     networks:
      - network
     command: postgres -c config_file=/etc/postgresql.conf


docker-compose.prod.yml
```diff
@@ -32,6 +32,7 @@ services:
      - database.prod.env
     expose:
      -  5432
+    user: postgres
```
maltezacharias, line 35: Won't work if there is no postgres user on your system:

```
➜  match4everything git:(easier-deploy) docker run -u doesneverexist hello-world 
docker: Error response from daemon: unable to find user doesneverexist: no matching entries in passwd file.
ERRO[0000] error waiting for container: context canceled 
```

I tried to get postgres to work well as non-root and I ran into problems all the time, ultimately I abandoned it, but I still think we should support this. Docker-compose.yml supports Env substitution (not from the database.prod.env file unfortunately) see https://docs.docker.com/compose/compose-file/#variable-substitution 

docker-compose up by default uses .env if present for this. We could use var substitution to set a default uid (0) and make it configurable via .env to another uid. We should also switch to using a volume for this, but I think none of those solutions fix the very common error I got:

Run the container once as root by accident and all your permissions will be screwed up. I think this happens with and without volumes. This has happened way too often for me and we should find a way to prevent that

maltezacharias (changes requested):

docker-compose.prod.yml
Baschdl: Are changing users really a problem for people which are not working on the Docker deployment?

maltezacharias: Well, I never got  it to run reliably and this PR shouldn't work on a clean system. I think we should make the user configurable and document how to set it up in a way that you can't easily screw it up. The postgres container docs contain some info on this as well under section `Arbitrary --user Notes` on https://hub.docker.com/_/postgres

Baschdl: > Won't work if there is no postgres user on your system

That shouldn't be a problem anymore with the `easier-deploy`, right?

Baschdl: @maltezacharias

maltezacharias: Unfortunately still is. In easier-deploy I left out the database container intentionally. I don't know whether the volume solved this because I haven't really investigated how docker handles the file permissions. The main problem is you would need a named postgres user on your docker host which you usually won't have, unless you installed postgres locally

Baschdl: Ok, I will just close this PR then

maltezacharias: To me the key to a solution for this would be that you can
- configure it to use a user you have setup on your system for the job
- make sure that you don't accidentally run the container as root, because that would reset permissions and cause problems later

I think a good solution for this is to create an .env (not to be confused with the env files we load) file where you can specify the host user accounts to use in an env variable and default to root uid if the env variables are not set. Docker auto-loads the .env file and you can then use env var substitution in the compose files. This would also be feasible for the backend. Although I still think gosu would be better there, I read through the security reports and think it would be ok to use, especially because we would still be better off than running as root. The .env approach would probably also work out as well, but running as a user without entries in /etc/passwd can have hard to debug sideeffects in the future.

maltezacharias:
@Baschdl I commented above, this is not yet resolved

Baschdl:
@maltezacharias Then please open an issue for it, this PR is far away from your proposed solution

PR #38, opened 2020-06-18
Revert "Delete code from match4healthcare #git"

feeds:
This reverts commit d4a162016b5617f0b5a0006f1fc87e0f6fa5781c.

We want to keep the old code to incentivize people to not do the same old thing again, but reuse what is there and reflect on what can be done better. This is easier when the code is just in the same repo.


bjrne (approved):
If you really really really really really want to add it back in :see_no_evil: 
Personally, I actually find it more difficult to work with, because a project wide search now automatically includes the old repository, which I have to manually exclude (when I remember to do so :wink: ).
And when comparing side to side in the same IDE, files have the same name and appear in the same window, when I would normally use two for comparing. But I guess, all is solvable somehow :woman_shrugging: :stuck_out_tongue_winking_eye:

Baschdl:
The project wide search is indeed a point a did not consider yet

feeds:
> The project wide search is indeed a point a did not consider yet

is that a disapprove? Otherwise I would merge this

bjrne:
I guess this was solved [here](https://github.com/match4everyone/match4everything/blob/383c5f8dd4b70b2b096756f5d2fe5cb0659022bb/.gitignore#L206), so i will close this for now.
https://github.com/match4everyone/match4everything/blob/383c5f8dd4b70b2b096756f5d2fe5cb0659022bb/.gitignore#L206

PR #48, opened 2020-06-20
Create a class-based config file for properties

feeds:
Design a class based config file like we talked over with @maltezacharias 
Generate a table for information about each participant based on this
Enable a signup with the required information and enable editing this info in the profile


feeds:
@josauder we need to somehow limit the length of the derived column names, as it is they can get too long and postgres will fail, while sqlite locally works

```
matching.ParticipantInfoB: (models.E018) Autogenerated column name too long for field 
"professional_training--medical_student_or_doctor--
recognition_as_an_internship_or_other_study_requirements_is_important".
 Maximum length is "63" for database "default".
```

josauder:
> @josauder we need to somehow limit the length of the derived column names, as it is they can get too long and postgres will fail, while sqlite locally works
> 
> ```
> matching.ParticipantInfoB: (models.E018) Autogenerated column name too long for field 
> "professional_training--medical_student_or_doctor--
> recognition_as_an_internship_or_other_study_requirements_is_important".
>  Maximum length is "63" for database "default".
> ```

I think it should be okay to simply display an error message in case the column names are too long.

Baschdl:
@josauder What would be the next step for the user when seeing the error message? Changing the A/B property names?

backend/apps/matching/configuration/properties.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def generate_random_assignment(self, rs=None):
+        raise NotImplementedError
+
+    def get_private_fields(self):
+        if self.private:
+            return [True for i in self.get_model_field_names()]
+        else:
+            if hasattr(self, "properties"):
+                return chain(*[p.get_private_fields() for p in self.properties])
```
Baschdl: How can a property have properties? Is `properties` semantically correct? Is this the nicest design?

backend/apps/matching/configuration/participant.py
```diff
@@ -0,0 +1,24 @@
+from itertools import chain
+
+
+class ParticipantConfig:
```
Baschdl, line 4: Only getters no setters?

backend/apps/matching/configuration/properties.py
```diff
@@ -0,0 +1,244 @@
...
+        self.info_text = info_text
+        self.private = private
+
+    def get_model_field_names(self, prefix=None):
+        raise NotImplementedError
+
+    def get_model_fields(self):
+        raise NotImplementedError
+
+    def generate_random_assignment(self, rs=None):
```
Baschdl: When do we want some randomness?

Baschdl: It's hard to review without a documentation string like the nice one for `__init__`. I will stop here for the moment.

Baschdl (dismissed):

backend/apps/matching/configuration/properties.py
feeds: its supposed to be a structure that can be iterated and group substructures like a tree. E.g. a `PropertyGroup` is grouping several properties. A conditional property has properties that are displayed only if a boolean property is selected.

I have not come up with a nicer design, and jonny as well

backend/apps/matching/configuration/properties.py
feeds: When we generate fake users for testing

backend/apps/matching/configuration/participant.py
feeds: no, all variable setting is done in the initialisation and then the variables are `final` in a java sense, so there is no need for setters.

backend/apps/matching/configuration/properties.py
feeds: I just pushed an updated version with more detailed comments. Could you try to review again.

Maybe start at the 'A.py' where the config is defined. I think this will make everything more clear.

maltezacharias (approved):
Source looks good to me, I tested signup for both A and B.
OK from me!

Issue #25, opened 2020-06-15
CMS system for use case-specific texts


Baschdl:
Users without programming knowledge should be able to change texts and forms

Specifics have to be discussed

Baschdl:
django cms plugins (user-editable content) could be a solution: http://docs.django-cms.org/en/latest/how_to/placeholders.html#placeholders-outside-cms

Baschdl:
Wagtail also supports our use case: https://docs.wagtail.io/en/latest/getting_started/integrating_into_django.html# and https://docs.wagtail.io/en/latest/advanced_topics/add_to_django_project.html

kevihiiin:
I think there are two approaches on how to use the CMS:
a) adding CMS to our application 
b) add our application to the CMS like a plugin

This could be a guide for b) using django CMS
http://docs.django-cms.org/en/latest/introduction/03-integrating_applications.html
And why to use plugins in django cms
http://docs.django-cms.org/en/latest/topics/plugins.html#why-would-you-need-to-write-a-plugin

kevihiiin:
Maybe available plugins are important too?

**Django CMS**
Commonly used
http://docs.django-cms.org/en/latest/topics/commonly_used_plugins.html
Marketplace
https://marketplace.django-cms.org/en/addons/

**Wagtail**
https://github.com/springload/awesome-wagtail

Baschdl:
Everything basically boils down to the question how much a non-html & non-python user and a non-python user should be able to edit

Baschdl:
https://pypi.org/project/django-template-editor/ would be an easy way to have an UI to change the templates

Baschdl:
Posted a question on SO, maybe the web has a better idea: https://stackoverflow.com/questions/62487635/combining-django-templates-with-content-editable-for-non-technical-users

Baschdl:
https://docs.djangoproject.com/en/dev/ref/contrib/flatpages/ could also be a way to achieve this

Baschdl:
Django-cms with custom plugins and [djangocms-snippet](https://github.com/divio/djangocms-snippet) for Javascript could work but would a complete rebuild of the design in django-cms require if I understand the workflow correctly. [djangocms-spa-vue-js](https://github.com/dreipol/djangocms-spa-vue-js) for vue.js single-page application support if we want this.

maltezacharias:
[...]

PR #49, opened 2020-06-20
Integrate django-cms into our project

Baschdl:
Part 1/n for #25: This PR only includes the setup, we have to integrate our sites separately into django-cms. Open question: how to automatically load the initial pages?

The user interface is available at [http://localhost:8000/admin](http://localhost:8000/admin) and you can login with a superuser


backend/requirements.txt
```diff
@@ -1,17 +1,33 @@
...
+djangocms-style==2.3.0
+djangocms-video==2.3.0
+djangocms_text_ckeditor==3.9.1
 djangorestframework==3.11.0
 djangorestframework-filters==0.11.1
 JSON-log-formatter==0.3.0
 numpy==1.18.2
 pandas==1.0.3
 pre-commit==2.2.0
+psycopg2==2.8.5
```
feeds: Do we really need this?

feeds (changes requested):
Otherwise the setup works for me. Should there anything else that should be achievable with this state other than the app not crashing?

feeds (approved):

backend/requirements.txt
Baschdl: Moved to production requirements

backend/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM ubuntu:18.04
 
-RUN apt-get update && apt-get install -y python3 python3-pip libpq-dev gettext
+RUN apt-get update && apt-get install -y python3 python3-pip libpq-dev gettext python-dev
```
maltezacharias, line 3: What is/was this change needed for?

Baschdl: I think one of the django-cms plugins needed it but I didn't try it without it. No idea why they would need python2.

PR #50, opened 2020-06-21
Create an approval interface

### LINKED ISSUE #35: Create an interface for approving participant A/B
We would like to enable staff to approve or disapprove participant A or B's, analogous to the m4e

Nice improvements would be:
- no pagination errors ;)

Not in scope: 
- viewing the details of a participant

feeds:
Closes #35

- no pagination errors anymore :)
- checks for approval permission (which is set for superusers by default. In order to test the behaviour you need to create a superuser and retract this right from him in the admin interface m4e.de/admin).
- created new issue #51 for creating a nice admin structure and deleting the useless staff user, that can be put on top of this (or we can merge that one first, however you prefer)

I decided to check the permission only when the change_approval button is clicked, so that other people without the permission can still view participants using this interface, but this is debatable - what do you think there?


bjrne:
I think greyed-out buttons with a help-text on hover would be the easier choice for users without the permission to approve, or removing the buttons altogether. The buttons should still also check for it though, since I guess you could fake a button press through the console somehow.

.gitignore
```diff
@@ -203,3 +203,4 @@ docker-compose.yml
 
 # Ignore all config files
 *.env
+/00_old_m4h_matching_code/
```
bjrne, line 206: Did you want to commit this here? :wink:

backend/apps/matching/migrations/0002_permission_group_creation.py
```diff
@@ -74,6 +74,8 @@ def create_groups(apps, schema_editor):
 
         group_is_a, created = Group.objects.get_or_create(name="is_a")
         group_is_b, created = Group.objects.get_or_create(name="is_b")
+        group_is_a_approved, created = Group.objects.get_or_create(name="approved_a")
+        group_is_b_approved, created = Group.objects.get_or_create(name="approved_b")
```
bjrne, line 78: @Baschdl strongly opposed changing existing migrations now, but I agree that this is the better solution here as several intertwined data and schema migrations would be very hard to squash later-on.

backend/apps/matching/models/participant.py
```diff
@@ -25,9 +25,10 @@ class AbstractParticipant(models.Model):
     # wants to take part in the matching process
     is_activated = models.BooleanField(default=True)
 
-    # valiFdation of account by staff
+    # validation of account by staff
```
bjrne, line 28: Maybe we should use a different term for "user, who has more permissions than others", since this technically is not a staff user here.

backend/apps/matching/models/participant.py
```diff
@@ -48,6 +49,29 @@ def new(cls, participant_type, email, **kwargs):
...
         return self.participant_type
 
+    @transaction.atomic
+    def change_approval(self, approver):
+        p_type = self.p_type().lower()
+
+        # check permission only here in case it was forgotten in the view
+        # maybe we can give a nicer message here?
+        if not approver.has_perm("matching.perm_approve_%s"):
+            raise ValueError("not allowed")
```
bjrne: I like this.

For the error message, maybe something like: "You currently don't have the permission to approve users. If you think this is a mistake, contact xyz." - but who would xyz be? A dynamic choice would be overkill I think, but not every page will have an imprint or a support mail adress...

backend/locale/de/LC_MESSAGES/django.po
```diff
@@ -344,6 +344,25 @@ msgid "Please validate your email by clicking this:"
...
+"\n"
+"    Approve\n"
+"    "
+msgstr  ""
+"\n"
+"    Freischalten\n"
+"    "
+
+msgid "These participants require approval."
+msgstr "Diese Teilnehmer bedürfen der Freischaltung."
```
bjrne: High society hm? :D how about
```suggestion
msgstr "Diese Teilnehmer müssen noch freigeschaltet werden."
```

bjrne: Or from another point of view
```suggestion
msgstr "Diese Teilnehmer wurden noch nicht freigeschaltet."
```
oder
```suggestion
msgstr "Diese Teilnehmer sind noch nicht freigeschaltet."
```

backend/locale/de/LC_MESSAGES/django.po
```diff
@@ -344,6 +344,25 @@ msgid "Please validate your email by clicking this:"
...
+
+msgid "These participants are approved, but you can disapprove them any time."
+msgstr ""
+"Diese Teilnehmer sind freigeschaltet, aber Sie können sie jederzeit ablehnen."
```
bjrne:
```suggestion
"Diese Teilnehmer sind freigeschaltet, aber Sie können die Freischaltung jederzeit wieder rückgängig machen."
```
Siezen wir auf deutsch?

bjrne (commented):
Will test locally

backend/apps/matching/models/participant.py
```diff
@@ -48,6 +49,29 @@ def new(cls, participant_type, email, **kwargs):
     def p_type(self):
         return self.participant_type
 
+    @transaction.atomic
+    def change_approval(self, approver):
```
bjrne, line 53: I am unsure whether I like a toggle-button for this. What happens when I press the button twice on slow PCs/internet?

backend/apps/matching/tables/approval.py
```diff
@@ -0,0 +1,53 @@
...
+    '  <div class="modal-dialog" role="document">'
+    '    <div class="modal-content">'
+    '      <div class="modal-header">'
+    '        <h5 class="modal-title" id="exampleModalLabel">Are you sure?</h5>'
+    '        <button type="button" class="close" data-dismiss="modal" aria-label="Close">'
+    '          <span aria-hidden="true">&times;</span>'
+    "        </button>"
+    "      </div>"
+    '      <div class="modal-body">'
+    "        Are you sure you want to delete the user with email {{ record.user__email }}?"
```
bjrne: This does not work unfortunately:

Screenshot from 2020-06-21 13-33-34

backend/apps/matching/tables/approval.py
```diff
@@ -0,0 +1,53 @@
+import django_tables2 as tables
+from django_tables2 import TemplateColumn
+
+from apps.matching.models import Participant
+from apps.matching.src.dual_factory import instanciate_for_participants
+
+approval_button = '<form>{% csrf_token %}<input type="hidden" value="{{record.uuid}}" name="uuid"> <button type="submit" name="change_approval" formmethod="post" class="btn btn-sm {% if record.is_approved %} btn-warning {% else %} btn-success {%endif%}">{% if record.is_approved %}Disapprove{% else %}  Approve {%endif%}</button></form>'
+info_button = (
+    '<a class="btn btn-info btn-sm" href="/matching/hospital_view/{{record.uuid }}">More Info </a>'
```
bjrne: This should redirect to a different link or include a TODO with a linked github issue

backend/locale/de/LC_MESSAGES/django.po
```diff
@@ -365,6 +387,14 @@ msgstr "Sie haben Ihr Profil aktiviert."
 msgid "Successfully updated info."
 msgstr "Sie haben Ihre Infos aktualisiert."
 
+#, python-brace-format
+msgid "You deleted the participant with email '{name}'."
+msgstr "Sie haben den Teilnehmer mit der E-Mail-Adresse '{name}' gelöscht."
+
+#, python-brace-format
+msgid "You approved the participant with email '{name}'."
```
bjrne: This is also used as the disaproval success message :see_no_evil:

bjrne (changes requested):
At least 
- [x] missing email in dialog
- [x] wrong disaprove message
- [x] view_hospital link
Have to be resolved, the rest are comments :)

bjrne:
Oh and one last thing: I got `Layout was forced before the page was fully loaded. If stylesheets are not yet loaded this may cause a flash of unstyled content.` in my browser console, and did indeed see the buttons flashing, but this likely has nothing to do with this PR. Maybe @maltezacharias knows what this stems from?

.gitignore
feeds: yes.

backend/apps/matching/models/participant.py
feeds: discussed this in #51

backend/apps/matching/models/participant.py
feeds: I added the error message as a raise by python.

backend/apps/matching/tables/approval.py
feeds: Fixed

backend/apps/matching/models/participant.py
feeds: nothing weird, the transactions are atomic so there should be no weird stuff hapenning

backend/locale/de/LC_MESSAGES/django.po
```diff
@@ -392,8 +392,8 @@ msgid "You deleted the participant with email '{name}'."
 msgstr "Sie haben den Teilnehmer mit der E-Mail-Adresse '{name}' gelöscht."
 
 #, python-brace-format
-msgid "You approved the participant with email '{name}'."
-msgstr "Sie haben dem Teilnehmer die E-Mail-Adresse '{name}' freigeschaltet."
+msgid "You changed the approval of the participant with email '{name}'."
+msgstr "Sie für den Teilnehmer mit der E-Mail-Adresse '{name}' den Status geändert."
```
bjrne:
```suggestion
msgstr "Sie haben für den Teilnehmer mit der E-Mail-Adresse '{name}' den Status geändert."
```

bjrne (approved):
Nice :beach_umbrella:

backend/apps/matching/migrations/0002_permission_group_creation.py
```diff
@@ -97,6 +99,8 @@ def delete_groups(apps, schema_editor):
         group_list = [
             "is_a",
             "is_b",
+            "approved_a",
+            "approved_b",
```
maltezacharias, line 103: Sorry, late to the party. Why two groups? Wouldn't the combination is_a + approved be enough?

backend/apps/matching/models/participant.py
```diff
@@ -48,6 +49,29 @@ def new(cls, participant_type, email, **kwargs):
...
+        p_type = self.p_type().lower()
+
+        # check permission only here in case it was forgotten in the view
+        # maybe we can give a nicer message here?
+        if not approver.has_perm("matching.perm_approve_%s"):
```
maltezacharias, line 58: Building permissions like this seems error prone. Also there seems to be a `% p_type` missing?

backend/apps/matching/models/participant.py
```diff
@@ -48,6 +49,29 @@ def new(cls, participant_type, email, **kwargs):
...
+            raise ValueError("You currently don't have the permission to approve users.")
```
maltezacharias, line 59: Maybe PermissionDenied Exception?
https://docs.djangoproject.com/en/3.0/ref/exceptions/#permissiondenied

backend/apps/matching/views/__init__.py
```diff
@@ -4,6 +4,7 @@
 from .participant_profile_edit import ParticipantInfoUpdateView  # noqa
 from .participant_signup import ParticipantSignup  # noqa
 from .profile_dashboard_redirect import ProfileDashboardRedirect  # noqa
+from .staff_approvals import ApproveParticipantsView  # noqa
```
maltezacharias, line 7: OOC: why is # noqa needed here?

feeds: bc we need to import these into the init if we want to make them available via `views.ApproveParticipants...`
and flake sees them as unused import, hence we need to ignore their warning

backend/apps/matching/migrations/0002_permission_group_creation.py
feeds: If we want to give the group of approved a's different permissions than b, it would become an issue
we do not test for groups, but for permissions that are assigned to the groups

backend/apps/matching/models/participant.py
feeds: did this in #64

backend/apps/matching/models/participant.py
feeds: yes, thanks :)

PR #52, opened 2020-06-21
Add option for osm tile server

### LINKED ISSUE #8: Mapbox vs OSM tile server
Should we add the public OSM tile server as an option and make it the standard setting? With this, a first installation wouldn't be dependent on Mapbox

feeds:
Closes #8


bjrne (changes requested):
I have not yet figured out why, but with and without docker, I don't get osm tiles. The browser instead requests mapbox tiles but without an api token. Feel free to dismiss this if it is my fault :sweat_smile:

backend.prod.env.example
```diff
@@ -4,5 +4,7 @@ SECRET_KEY=
 SENDGRID_API_KEY=
 # Slack Logging Webhook
 SLACK_LOG_WEBHOOK=
-# Token from mapbox.com to get Open Street Map tiles
+# Mapview Backend ('open_street_map' or 'mapbox')
+MAPVIEW_BACKEND=
```
Baschdl:
```suggestion
LEAFLET_TILESERVER=
```

backend/apps/checks.py
```diff
@@ -92,19 +93,42 @@ def check_slack_webhook(app_configs=None, **kwargs):
...
+def check_map_settings(app_configs=None, **kwargs):
     errors = []
 
-    if settings.MAPBOX_TOKEN is None:
+    if settings.MAPVIEW_BACKEND == "open_street_map":
+        errors.append(
+            Warning(
+                "Usage restricted tile server.",
+                hint="You are using an open street map tile server for viewing the map, which is subject to usage "
+                "restrictions. \nIf you plan to put this system into production, please consider setting up "
```
Baschdl:
```suggestion
                "restrictions (See https://operations.osmfoundation.org/policies/tiles/). \nIf you plan to put this system into production, please consider setting up "
```

backend/apps/checks.py
```diff
@@ -92,19 +93,42 @@ def check_slack_webhook(app_configs=None, **kwargs):
...
+            )
+        )
+    elif settings.MAPVIEW_BACKEND == "mapbox":
+        if settings.MAPBOX_TOKEN is None:
+            errors.append(
+                Error(
+                    "Mapbox token not found.",
+                    hint=(
+                        "You have to set the Mapbox token in you environment with "
+                        "'export MAPBOX_TOKEN=<<yourToken>>'."
```
Baschdl:
```suggestion
                        "'export MAPBOX_TOKEN=<<yourToken>>'. or if you use Docker in the corresponding env file"
```

Baschdl (dismissed):

backend/apps/checks.py
maltezacharias: Either everywhere or nowhere. I'd suggest changing the wording to "set the environment Variable MAPBOX_TOKEN"

maltezacharias: I mean we should change this in all checks and remove "export ....=...." because that's bash specific. It won't work neither with docker nor Windows env for example. We should keep the error message environment-agnostic.

maltezacharias (commented):
I don't know why Github diffs show the entire mapview.js file as modified, 
in my local env only a few lines are changed. Any idea why it is like that?

I think it might be a better idea to just directly pass the tileLayerURL via Env. Same goes for Mapattribution. That way no source changes would be needed for different map providers. I see no benefit in adding coding for each mapprovider if it's just those two lines that are affected.

Is there anything I'm missing?

feeds:
we could also offer the option "tile server url" and let people set up that if they care. The advantage of the two options is that people do not need to research what a tile server is and that we can have mapbox specific logic to validate that we have an api key present.

backend/static/js/mapview.js
```diff
@@ -1,172 +1,172 @@
...
+
+
+    initializeMap: function initializeMap() {
+        let mapOptions = {
+            center: [51.13, 10.018],
+            zoom: 6
+        }
+        let tileLayerURL = this.options.tileURL
+        let tileLayerOptions = {
+            attribution:  this.options.mapAttribution + '<a href="https://www.mapbox.com/map-feedback/" target="_blank">Improve this map</a> | Icons by <a href="https://www.flaticon.com/authors/freepik" title="Freepik">Freepik</a> from <a href="https://www.flaticon.com/" title="Flaticon">flaticon.com</a>',
```
feeds, line 70: @maltezacharias the attribution links that are contained in mapAttribution are rendered to text here (lower corner of the map). Any idea on how to fix this?

backend.prod.env.example
feeds: Yep

backend/apps/checks.py
feeds: Removed all exports as well

Baschdl:
> in my local env only a few lines are changed

@maltezacharias You have one guess what the cause is 😆 Hint: The line ending matches yours and not Freya's

backend/apps/checks.py
Baschdl: @maltezacharias @feeds I would mention it in the help texts because the only reason for the texts is ease of use and it's not obvious that "set it in your environment" doesn't have any effect when someone uses Docker (which is (with an easy setup from our side) the most likely option if someone just wants to try it out)

bjrne:
> @maltezacharias You have one guess what the cause is 😆 Hint: The line ending matches yours and not Freya's

But with the commit hooks, the line endings should be uniform 🤔

bjrne:
Ah, mystery solved. File was in crlf mode, @feeds did not have pre-commit installed when pushing the new repo, github actions were not yet activated and thus this thing got online. so now it is fixed ;)

bjrne (changes requested):
I have no idea how, but we need to remove the mapbox logo when we're not using their map (and in theory, I think it should be bigger in the mapbox case).
Screenshot from 2020-06-25 15-10-45
Crazily, osm now works like a charm on my machine, and I like the approach taken with the env variables. so this is basically an approve with one more thing ;)

feeds:
@bjrne can you maybe open a seperate issue for that?

its probably js I do not know how to handle

bjrne:
#59 is open.

bjrne (approved):

backend/static/js/mapview.js
maltezacharias: see map.html

backend/apps/matching/templates/map/map.html
```diff
@@ -26,8 +26,8 @@
     bURL     : "{% url 'participant_JSON' 'B' %}",
     aListURL  : "{% url 'participant_list' 'A' 'COUNTRYCODE' 'PLZ' 0 %}",
     bListURL   : "{% url 'participant_list' 'B' 'COUNTRYCODE' 'PLZ' 0 %}",
-    tileserver: "{{ tileserver }}",
-    mapboxToken       : "{{ mapbox_token }}",
+    tileURL: "{{ tileserver }}",
+    mapAttribution : `{{ attribution }}`,
```
maltezacharias: https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#safe
Need to add `| safe` filter

maltezacharias (changes requested):

maltezacharias:
For me this looks good except for the escaping on map.html

backend/apps/matching/templates/map/map.html
feeds: thanks :)

backend/static/js/mapview.js
feeds: Done

backend/apps/checks.py
feeds: I'm not sure I understand this correctly. Can you maybe make a commit suggestion?

maltezacharias (approved):

Issue #42, opened 2020-06-19
Timezone handling


bjrne:
We currently sometimes use utc times, sometimes localized times: we need to draft a coherent approach for the project.
Notes:
- There is no standard like the request-language http header for timezone localization, thus we need to think about how to handle this. A classic approach is to infer time-zones from languages and countries or IP-Address ranges. However given our principle of data economy, this is not my favored approach

- If we have it, use the postal code to save a hidden timezone field to user profile. This could also appear as editable on the user profile page. This could be a difficult route for customization later-on: what happens if the use-case does not collect postal codes?

- If we don't, we should think about a good enough fallback solution. A timezone picker next to datetime fields? A standard model field that defaults to GMT or a sensible heuristic?

maltezacharias:
1. Store UTC everywhere
2. See if inferring from ZIP code is possible, alternatively set default timezone in config
3. Profit
4. Or add to user profile and make it changeable. Considerung that we should be able to show the users sent messages and an inbox/outbox maybe we need to be able to handle timezones.

Baschdl:
Do we need the time for anything except internal analysis?

bjrne:
Availability checking in the matching process. If the matching is local, they should be in the same timezone, but if the matching is for online tutoring, IT-support, ..., timezone support becomes critical.

maltezacharias:
Inbox

PR #53, opened 2020-06-21
Fix naive datetime warning

### LINKED ISSUE #34: Fix naive datetime warnings
Does not include full timezone support, this is will be part of #42.

bjrne:
Closes #34 

We currently do not fully support timezones and therefore the flag `USE_TZ` is misleading and generated warnings. Full timezone support will be added in #42.


josauder (approved):
Looks good to me

backend/apps/matching/management/commands/createfakeusers.py
```diff
@@ -79,7 +80,7 @@ def delete_all_fakes(self):
     def add_fake(self, participant_type, n):
         n_users = User.objects.all().count()
 
-        for i in range(n):
+        for i in tqdm(range(n)):
```
Baschdl, line 83: Interesting connection to time zones 🤣

Issue #56, opened 2020-06-23
Add a validator for emails


feeds:
in the signup form, just like we had in m4h

josauder:
![image](https://user-images.githubusercontent.com/12648328/86130056-e3c67c80-bae3-11ea-9f78-5865d250d226.png)


I think this is actually currently implemented. Feel free to close if this is what you mean

feeds:
Oh, right :)

PR #63, opened 2020-06-26
Enable adding several locations to one profile

feeds:
image

requires #56 (for consistency)


josauder (approved):
Works great, and looks good to me

PR #65, opened 2020-06-26
Add a persistent filter to our model

feeds:
Create a new Filter that is saved in `http://127.0.0.1:8000/matching/A/filter/create`

or call the rest api to filter the existing participants with get parameters `http://127.0.0.1:8000/matching/api/A/info/list` (see  OPTIONs-button)


feeds:
@bjrne can you explain why the tests are failing? I have no idea.. :grin:

feeds:
@maltezacharias This is what I should you ... what else do you need?  :)

backend/locale/de/LC_MESSAGES/django.po
```diff
@@ -18,6 +18,18 @@ msgstr ""
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 "X-Generator: Poedit 2.3.1\n"
 
+msgid "is exactly"
+msgstr ""
```
bjrne: I guess because of this :D

bjrne (dismissed):

feeds:
Use #74 instead

Issue #69, opened 2020-07-01
Include config for B, script for rebuilding project from configuration.


josauder:
- As of right now, A and B both use the class for A.
- Also, building the project right now from scratch requires moving `0002_permission_group_creation.py` to a different folder, deleting all other migrations, running `makemigrations && migrate`, and then re-adding `0002_permission_group_creation.py`.

bjrne:
Now that we use the automatic configuration, there might be a better way to add the groups, like automatically loading a fixture? I'm not a huge fan of it since it is less verbose in code, but it could be way easier here.

Baschdl:
Closed with #70

PR #70, opened 2020-07-01
69 config for b rebuild script

josauder:
solves #69


scripts/rebuild_from_configuration.sh
```diff
@@ -0,0 +1,9 @@
+rm run/db.sqlite3
+mkdir ___tmp
+cp apps/matching/migrations/0002_permission_group_creation.py ___tmp
+rm apps/matching/migrations/0*
+python3 manage.py makemigrations
+python3 manage.py migrate
```
bjrne: This could be removed I think, but it might as well stay

bjrne (approved):

Baschdl:
Depends on #124

PR #74, opened 2020-07-01
Create persistent filters with dashboard and contact possiblities

feeds:
Create a new Filter that is saved in `http://127.0.0.1:8000/matching/A/filter/create`

or call the rest api to filter the existing participants with get parameters `http://127.0.0.1:8000/matching/api/A/info/list` (see  OPTIONs-button)


feeds:
Duplicate

Issue #37, opened 2020-06-17
Create a persistent filter that can be applied to a participant set


feeds:
We want to be able to save a filter over the information of a participant, so we can follow up on this at a later point in time (e.g. if participants were added post-filtercreation that match the filter) and the participant can save the search in their dashboard.

This includes

- [ ] creating a custom filter for the location and time info
- [ ] creating a filter that is generated from the participant info (it might be configurable, but it should not require a config, maybe as a later issue and for now use defaults). This could mean that a text property could allow open text search, bool fields have yes/no/none, choices can be multiselected and ordered choices can be range selected etc. Some of that functionality could be reused from `django-filter`. 

#### Open Question
Do we allow more intricate logic to compose the filters (and, or...)? Facebooks search graph was an experiment that allowed this but I guess most of the population never used the features that the developers created.... ;)

Baschdl:
"more intricate logic" would be cool but we would need to design it in a way that it can be understood easily. An idea could be to have it very low-key on the UI, only "power user" use it and the rest doesn't really notice it. How to implement it without building an interface for SQL is another question.

PR #81, opened 2020-07-04
Javascript filtering UI using Vue.js

### LINKED ISSUE #30: Create a filterable list overview for the participants over each other
Depends on #37 
What technique do we want to use here?

### A)
- just plain old django filtering as before, with a page reload when the filter is updated (note that github does this for issue filtering and it is hardly noticable), but with a custom filter design.

### B)
- a frontend that calls
-  our [rest api](https://www.django-rest-framework.org/api-guide/filtering/) that loads the appropriate participants on demand without necessitating a complete page reload, which seems to be quite easy
- we should also provide an endpoint to the frontend that gives back all the filtering options, so this also is configurable, but here we need to discuss the format

## Other resources
- [js table generator for filtering that is kinda cool :)](https://mdbootstrap.com/docs/jquery/tables/generator/)
- [table design suggestions](https://semantic-ui.com/collections/table.html)

### Open questions
- could we implement the hierachical enabling of filters in this way, like we had for m4h? (only enable age filter for students if they are required?)

maltezacharias:
Based on top of #37 & #18 

- [X] Integrate Vue.js
- [X] Create a PoC for a Vue Component to select  a filter value
- [ ] Build a filter form from a mock up json definition file
- [ ] Retrieve and display results for above filter form
- [ ] Retrieve filter definition .json from backend and build filter from this


maltezacharias:
Will integrate into #173

Baschdl:
@maltezacharias Wouldn't a second PR be easier to review?

PR #95, opened 2020-07-08
Show mapbox attribution and logo only if mapbox is used

### LINKED ISSUE #59: Remove map box icon when using other tileservers
Currently,the mapbox watermark appears in the bottom left corner of the map regardless of whether the map box api is used.

Scope of this issue is to hide it when another tileserver url is used.

Baschdl:
Closes #59

Depends on #21

### DIFF
--- a/backend/static/js/mapview.js
+++ b/backend/static/js/mapview.js
@@ -8,6 +8,7 @@ mapViewPage = {
         aListURL  : '',
         bListURL   : '',
         tileURL: '',
+        tileServer: '',
         mapAttribution: '',
         isStudent: true,
         isHospital: true,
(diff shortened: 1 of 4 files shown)


backend/static/js/mapview.js
```diff
@@ -67,13 +68,18 @@ mapViewPage = {
...
             id: 'mapbox/streets-v11',
             tileSize: 512,
             zoomOffset: -1,
             preferCanvas: true,
-          }
+        }
+        if (this.options.tileServer == 'mapbox') {
+          tileLayerOptions.attribution += '| <a href="https://www.mapbox.com/map-feedback/" target="_blank">Improve this map</a> ';
+        }
+        tileLayerOptions.attribution += '| Icons by <a href="https://www.flaticon.com/authors/freepik" title="Freepik">Freepik</a> from <a href="https://www.flaticon.com/" title="Flaticon">flaticon.com</a>';
```
bjrne: These icons are not used anymore, they were added back in efd1d6a and looked like this:
Screenshot from 2020-07-08 23-27-35
but are not used anymore, so this can be removed.

bjrne (approved):
Otherwise good to go :bee:

PR #92, opened 2020-07-08
Use django-cms for most pages

Baschdl:
Provide configuration for pages in django-cms

Depends on #21 
Fixes  #87 and fixes #25

PR #96, opened 2020-07-08
Describe loading the initial data

### LINKED ISSUE #88: Add initial loading of page setup to docker setup
After creating a initial setup in #87, add automatic initial loading of this data

Baschdl:
In contrast to #88, loading the initial data should be an active action and not be executed by the automatic setup because it should only performed once.

Closes #88

Depends on #92


README.md
```diff
@@ -17,6 +17,9 @@ TODO
 We use the content management system [django-cms](https://www.django-cms.org/) to make the content editable by users which do not have a computer science background.
 The admin interface is accessible at `http://localhost:8000/admin/` with a superuser account.
 
+
+Dump is generated by `python3 manage.py dumpdata cms djangocms_text_ckeditor > cms_export.json` and loaded by `python3 manage.py loaddata`.
```
bjrne, line 22: I know it is kind of duplication, but since we already do it with every other command: Should we add the docker version of this as well?

bjrne (approved):
Otherwise good

README.md
maltezacharias: I'd much rather just once and for all write that in big letters at the start to create an alias for whatever you want to use:
```
For production docker:
alias django-admin=docker-compose -f docker-compose.yml -f docker-compose.prod.yml exec backend django-admin

For development docker:
alias django-admin=docker-compose exec backend django-admin

For local development just use django-admin and be sure to set PYTHONPATH to include backend and DJANGO_SETTINGS_MODULE to either match4everyone.settings.development or production
```
and then use django-admin everywhere in the documentation. As of now django-admin consistently works in all docker-environments we supply.

maltezacharias: I'd much rather just once and for all write that in big letters at the start to create an alias for whatever you want to use:
```
For production docker:
alias django-admin=docker-compose -f docker-compose.yml -f docker-compose.prod.yml exec backend django-admin

For development docker:
alias django-admin=docker-compose exec backend django-admin

For local development just use django-admin and be sure to set PYTHONPATH to include backend and DJANGO_SETTINGS_MODULE to either match4everyone.settings.development or production
```
and then use django-admin everywhere in the documentation. As of now django-admin consistently works in all docker-environments we supply.

maltezacharias (changes requested):

README.md
bjrne: Sounds good for the upcoming documentation ;)

bjrne: Since this is a general thing, I think this would be better to do this in the documentation and merge this as is. But I like the idea.

maltezacharias (approved):
Okay, @bjrne is right, we can do that in the documentation. But then I'd just ignore the whole comment about writing the docker command as well.

PR #97, opened 2020-07-08
Fix blurry osm tiles

### LINKED ISSUE #94: OpenStreetMap tiles are blurry
OpenStreetMap are displayed blurry, possibly a problem with an unnecessary projection.
image

Baschdl:
Use mapbox config only when mapbox is chosen as map provider

Closes #94 
Depends on #95


bjrne (approved):
Works :+1:

PR #58, opened 2020-06-23
Extend time until timeout in production

Baschdl:
Some changes orthogonal to our code (didn't we freeze all necessary package versions?) cause load times of more than 30 seconds for the first request (at least on my machine) and therefore cause the gunicorn worker to timeout.

Should fixes timeouts in #49 and possibly also error 500 in #21

PR #98, opened 2020-07-08
Use psycogreen

Baschdl:
> As per https://docs.gunicorn.org/en/stable/settings.html#timeout:
> 
> > Generally set to thirty seconds. Only set this noticeably higher if you’re sure of the repercussions for sync workers. For the non sync workers it just means that the worker process is still communicating and is not tied to the length of time required to handle a single request.
> 
> I did some reading up on the different worker_classes and found this:
> 
> > For full greenlet support applications might need to be adapted. When using, e.g., Gevent and Psycopg it makes sense to ensure psycogreen is installed and setup.
> 
> https://docs.gunicorn.org/en/stable/design.html
> 
> I think with async workers normally you shouldn't see a timeout happening here, have we implemented the above or is it possible that our DB calls are still synchronous? At least in requirements we have no psycogreen (yet?)

For more details see #58


maltezacharias:
It seems to work, that is really all I can say :D

maltezacharias (approved):

backend/requirements.prod.txt
```diff
@@ -1,5 +1,6 @@
 gevent==1.5.0
 gunicorn==20.0.4
 psutil==5.7.0
+psycogreen==1.0.2
 psycopg2==2.8.5
 psycopg2==2.8.5
```
bjrne, line 6: Was this a weird merge?

bjrne: @Baschdl

Baschdl: This one obviously only added `psycogreen==1.0.2` but the duplication of `psycopg` is indedd strange

PR #105, opened 2020-07-09
Fix deletion and deactivation/reactivation

### LINKED ISSUE #76: Handle deletion and deactivation of users
The links are in the dashboard, but do not work.

Best idea: Take what is there in m4h and copy+paste :)

Baschdl:
@feeds Are the `/`s behind `change_activation/`, `delete_me/` etc. intended? Is the `<p:p_type>` even when we don't need it intended or should we directly call `/delete_me_ask/`?

Fixes #76


backend/apps/matching/templates/messages/deleted_user_ask.html
```diff
@@ -46,9 +46,9 @@
   <div class="card login-card text-center align-items-center">
       <h4>{% blocktrans %}Are you sure you want to delete your account?{% endblocktrans %}</h4>
 
-      <a href="delete_me"><button class="btn blue text-white btn-md" style="margin-top:10px;">{% blocktrans %}Yes, delete my account.{% endblocktrans %}</button></a>
+      <a href="../delete_me"><button class="btn blue text-white btn-md" style="margin-top:10px;">{% blocktrans %}Yes, delete my account.{% endblocktrans %}</button></a>
```
maltezacharias: Why not url template tags?

maltezacharias: @feeds ?

maltezacharias (dismissed):

backend/apps/matching/templates/messages/deleted_user_ask.html
bjrne: I fixed it for all links :)

bjrne (approved):
Tested locally, works :)

bjrne (commented):
Although when I think about it again, let's just remove the p_type here altogether? For
- delete me
- activation links
- profile (Already in #126 )

EDIT: Added an issue for that, #148. Will merge this one as it is.

PR #106, opened 2020-07-09
Link the participant view in the approvals

### LINKED ISSUE #55: Link the participant view in the approvals
Approvers should be able to quickly get an overview over the participant, incuding "private" info like phone number to decide on the approval

Baschdl:
@bjrne is implementing the permissions

@feeds Did you design the urls deliberately like this?

Fixes #55


maltezacharias:
I commented on the original issue

feeds:
me too, there is a one liner to also see the admin fields

feeds:
I don't know what you are asking for @Baschdl 

> Did you design the urls deliberately like this?

Baschdl:
@feeds We already discussed the urls somewhere else, so that's fine

feeds (commented):
There is a nicer way to do this and I added this accidentally in #159, so this PR is no longer needed.

We need to figure out the permission system though (right now, everyone can only view themselves, #120), so I'll leave the issues open.

PR #108, opened 2020-07-09
Prevent User creation during login

### LINKED ISSUE #6: Prevent User creation during login
In general,  we want to prevent users from creating another account when they are logged in.
How do we prevent this?

Using the Permissions from django.auth?

Baschdl:
I was unable to write an `invert()` method for decorators, so here is a new decorator

Fixes #6


backend/apps/matching/admin.py
```diff
@@ -42,3 +42,15 @@ def new_func(request, p_type, *args, **kwargs):
     if function:
         return actual_decorator(function)
     return actual_decorator
+
+
+def no_login_required(function=None, login_url="/"):
```
bjrne: Maybe rather this?
```suggestion
def logged_in_not_permitted(function=None, login_url="/"):
```
`no login required` suggests it is not required, not that it is not allowed...

bjrne (approved):

PR #111, opened 2020-07-10
Cleanup

bjrne:
- Merge src and utils folder
- Name configuration folders consistently


backend/apps/matching/models/participant_info_location.py
```diff
@@ -7,7 +7,7 @@
 import numpy as np
 
 from apps.matching.files.map_data import plzs
-from apps.matching.src.plzs import BIG_CITY_PLZS
+from apps.matching.utils.plzs import BIG_CITY_PLZS
```
maltezacharias: as we're cleaning this up, suggest renaming to zipcode? Should we use singular or plural here?

Baschdl (approved):
Fine after `zipcodes`

bjrne:
I changed zipcodes everywhere except for in the models, I will clean that up after #70 is merged. Issue is #133.

Baschdl (approved):

PR #120, opened 2020-07-10
Fix permission problem

bjrne:


backend/apps/matching/views/participant_info_edit.py
```diff
@@ -18,6 +19,11 @@ class ParticipantInfoUpdateView(UpdateView):
     slug_url_kwarg = "uuid"
     slug_field = "uuid"
 
+    def dispatch(self, *args, **kwargs):
+        if self.request.user.participant().info.uuid not in self.request.path:
```
bjrne: This is not the nicest solution, but I could not think of a better one that works without a db-lookup. The url contains the uuid of the `participant_info`, not the participant or the user.

backend/apps/matching/views/participant_info_view.py
```diff
@@ -20,6 +21,11 @@ class ParticipantInfoViewView(UpdateView):
     slug_url_kwarg = "uuid"
     slug_field = "uuid"
 
+    def dispatch(self, *args, **kwargs):
+        if self.request.user.participant().info.uuid not in self.request.path:
```
Baschdl: In my understanding, hospitals should also be able to view it or we should at least have a toggle for this in the configurations

Baschdl:
@feeds Does `request.user.participant()` also work in a decorator?

feeds:
yes, it should work in a decorator.

Some remarks: 
Remember that the class-based view is in the end translated back into a function. I.e. everything that gets passed to dispatch can also be accessed in a method decorator to a class based view. i.e. `*args, **kwargs` where` kwargs['p_type']` gives the url p_type and `kwargs['uuid']` gives the uuid, so you can check in the method decorator for the permission via `ParticipantInfo[p_type].objects.get_or_404(uuid=uuid)` [(described here)](https://docs.djangoproject.com/en/3.0/topics/http/shortcuts/#get-object-or-404) and use django orm to navigate to the user `ParticipantInfo[p_type].objects.get_or_404(uuid=uuid).participant.user` and check whether it is the one from the request.

backend/apps/matching/views/participant_info_view.py
feeds: yes, this should definitely be configurable, after all, it would be pretty useless for a hospital if they could not view the complete info of a participant

backend/apps/matching/views/participant_info_edit.py
feeds: I would do it in reverse, as described in my comment below

feeds:
Oh but of course writing that decorator gets a bit more complicated because you need to pass the arguments correctly. I think putting it in dispatch is neat, especially if the config gets included so people can read the code more easily

bjrne:
In order to not do things two times, this now depends on #183 to see whether permissions change their names then.

bjrne:
Now depends on #200 .

feeds (approved):

PR #123, opened 2020-07-10
Only show buttons for sites a staff user has access to

Baschdl:
The staff interface should only show buttons for sites a staff user has access to

@bjrne will create `can_send_newsletter`


bjrne:
This does not really make sense to fix here, so now it depends on #124.

Baschdl (commented):
@bjrne's changes look good

feeds (approved):
I just fixed a small issue with the permission formulation. Otherwise everything works nicely.

PR #124, opened 2020-07-10
Add only non-configurable models in initial migration

bjrne:
Add groups and permissions in second one. Run 'makemigrations matching' to create third migration for the configurable tables. Configuration changes can now be added by deleting the third migration and running makemigrations.

Adding them as a new migrations only works if defaults are set or if the field is nullable. Weirdly enough, setting defaults is disabled currently for non-nullable fields, where they would be especially important. Maybe @feeds can elaborate? The code location is here, where `default=self.default` is left out:
https://github.com/match4everyone/match4everything/blob/e9578f184bab4720422d7e48292e5f48883ba01d/backend/apps/matching/configuration/properties.py#L242
and here:
https://github.com/match4everyone/match4everything/blob/e9578f184bab4720422d7e48292e5f48883ba01d/backend/apps/matching/configuration/properties.py#L292


feeds:
If we leave default out, a direct consequence is that these fields are automatically set as 'required' in the forms  that django creates (with the star, and automatic promts if there is no input). So whereever the config states the field as required, I do not want to set a default value.

feeds (approved):
Nice :)

Issue #125, opened 2020-07-10
matching_participant_required restriction of ParticipantDashboard


Baschdl:
@feeds What was the idea behind the `matching_participant_required` restriction here: https://github.com/match4everyone/match4everything/blob/4f4687cdfb512fda8e9f5bee6d9bb851566c27c7/backend/apps/matching/views/participant_dashboard.py#L16 
I don't see how this adds any security.

feeds:
It doesn't, I approved the PR

Issue #113, opened 2020-07-10
Clean up links


Baschdl:
Is there a greater concept behind our urls?
Currently, `A` and `B` is in our link. Do we want to dynamically use `ParticipantConfig.name` in the end?

Blocked by @feeds

feeds:
I think it would be cool to use the name form the config. But we should probably use a url_name variable in case the actual name co taines spaces ect

feeds:
and the greater concept... i intended everything that is non user-only specific to have the A/B binding

One inconsistency in the current implementation is that the profile page is A/B specific. That is non really neccessary since that page is not accessedby someone else

Baschdl:
I started in https://github.com/match4everyone/match4everything/tree/name_in_url but you probably have the better overview where `A` and `B` is used

PR #126, opened 2020-07-10
Resolve inconsistency where profile was dependent on p_type A/B

Baschdl:
Discussed in #113

Dependent on #125


feeds (approved):
This is exactly what I imagined :)

PR #131, opened 2020-07-10
Use prettier rtfd theme

Baschdl:
Separate requirements.txt is also used by readthedocs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,8 @@ COPY backend/requirements.txt /backend/requirements.txt
 RUN pip3 install -r requirements.txt
 COPY backend/requirements.prod.txt /backend/requirements.prod.txt
 RUN pip3 install -r requirements.prod.txt
+COPY docs/requirements.readthedocs.txt /backend/requirements.readthedocs.txt
+RUN pip3 install -r requirements.readthedocs.txt
 
 # Copy the JS Bundle from the intermediate stage over into the backend container
 # Important: Do this before collecting static, otherwise the bundles won't be found
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,6 +6,7 @@
 !/.dockerignore
 !/backend/
 !/frontend/
+!/docs/
 
 # Remove volatile directories for repeatable builds
 **/dist/*
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -30,7 +30,6 @@ pre-commit==2.2.0
 requests==2.23.0
 sendgrid==6.2.0
 six==1.15.0
-sphinx==3.1.2
 tqdm==4.46.1
 whitenoise==5.0.1
 whitenoise[brotli]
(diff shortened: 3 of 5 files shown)


bjrne (changes requested):
Throws errors:
```bash
$ make html
Running Sphinx v3.1.2

Exception occurred:
  File "/usr/lib/python3.8/importlib/__init__.py", line 118, in import_module
    if name.startswith('.'):
AttributeError: module 'sphinx_rtd_theme' has no attribute 'startswith'
The full traceback has been saved in /tmp/sphinx-err-bou_3p5h.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [Makefile:20: html] Error 2
```

bjrne (approved):

Dockerfile
```diff
@@ -27,6 +27,8 @@ COPY backend/requirements.txt /backend/requirements.txt
 RUN pip3 install -r requirements.txt
 COPY backend/requirements.prod.txt /backend/requirements.prod.txt
 RUN pip3 install -r requirements.prod.txt
+COPY docs/requirements.readthedocs.txt /backend/requirements.readthedocs.txt
```
bjrne, line 30: Oh no, you really needed to go to sleep :see_no_evil:

bjrne (changes requested):

Dockerfile
Baschdl: Oh no, you really need to learn Docker 🙈

Baschdl:
@maltezacharias Concerning https://github.com/match4everyone/match4everything/pull/131/commits/a26eb9f08bc334613f32b827cde92387fff896e5 and the current implementation of the `.dockerignore`, I have the feeling we introduced quite some complexity by design which we can document to make it easier but I find it questionable if the person which sees an error because of this reads our hint in the right moment. Why don't we ignore everything we don't want but by default include everything? I think that would make it easier for external users which would probably only add a toplevel folder once in a lifetime 😉

bjrne (approved):

maltezacharias:
Best practice is to keep the docker build context as small as possible and if we go the "only exclude some directories way" that would mean that you will always have to remember to exclude every new file you add that won't be needed for the docker build and I'm pretty sure that's not gonna happen. 

Using the "exclude all, but ..." approach means that your Docker build will fail unless you adressed the file -> Keeping the build context down to only the required files. The .dockerignore right now is already a mix to make errors less likely: "Include all of frontend and backend except for node_modules, pycache and ..." You should only need to adapt the .dockerignore when you change the architecture of the docker containers, e.g. introducing a new intermediate build container or something like that. We could add a hint to the Dockerfile to check the .dockerignore. Then again that is pretty basic Docker knowledge. I feel that is not needed but I won't be against it. I will also gladly write a section in the docs detailing the Docker build design and the design choices that went in there.

That is also no guarantee as @bjrne seems to have included the whole docs dir without using it at all. Please revert that change or explain why it was necessary.

maltezacharias:
EDIT @Baschdl added docs, I read the comments wrong. Still why?

PR #132, opened 2020-07-11
Show mail address in profile

### LINKED ISSUE #119: Show user email and username in profile
As a normal user of the system, I want to check what email I used in my profile information.

Baschdl:
What do our design-minded people think about this?
image

Closes #119


bjrne:
Since we do not offer the option to change the email address, I would prefer it if the field looks less editable :) Until we add that option of course...

bjrne:
Will review when cms fixture is fixed.

backend/apps/matching/templates/participant/participant_dashboard.html
```diff
@@ -41,7 +41,11 @@
                 {% blocktrans %}Profile Details{% endblocktrans %}
             </div>
             <div class="card-body">
-
+                <label for="email">{% blocktrans %}E-Mail{% endblocktrans %}:</label>
+                <div class="">
+                    <input type="text" id="email" name="email" value="{{ request.user.email }}" class="textinput textInput form-control" disabled="">
+                    <small id="email_change_hint_text" class="form-text text-muted">{% blocktrans %}Your account is irreversibly linked to your email address. Should you wish to use another one, please delete your account and register with a new one.{% endblocktrans %}</small>
```
bjrne, line 47: I added a small helptext whose design is copied from the crispy help texts.
According to [this](https://developer.paciellogroup.com/blog/2012/01/html5-accessibility-chops-title-attribute-use-and-abuse/) website, using the `title`-attribute for hover texts is bad for accessibility and touch devices.
The result looks like this:
Screenshot from 2020-07-12 00-15-55

bjrne (approved):
I like it now

backend/apps/matching/templates/participant/participant_dashboard.html
Baschdl: I like it

PR #138, opened 2020-07-11
Permission checks for approvement pages

### LINKED ISSUE #121: Only approver should be able to approve
Only user with 'can_approve_type_a' or 'can_approve_type_b' should see the corresponding approval interfaces and be able to approve someone. Currently, we only check for `is_staff`

https://github.com/match4everyone/match4everything/blob/ac991347e22c99b2ca91e181adc1bd8dde9be43c/backend/apps/matching/views/staff_approvals.py#L19

### LINKED ISSUE #137: Add permission check on user deletion
Currently, any staff user can delete users without a permission check. We should add the pre-generated django permission for this.

bjrne:
Closes #121 
Closes #137


backend/apps/matching/views/staff_approvals.py
```diff
@@ -21,6 +22,13 @@ class ApproveParticipantsView(View):
     template_name = "approve_hospitals.html"
 
     def get(self, request, p_type):
+        # If you are not a staff user, the method decorator takes care of showing a 404 page,
```
Baschdl: I don't get a 404 as fake user A1:
image

Baschdl (changes requested):

backend/apps/matching/views/staff_approvals.py
bjrne: Good catch - the django decorator `staff_member_required` redirects to the login page under the assumption that you need to log in again with the higher ranked user.

You can override this behaviour like I did in cc5057d and 892abf0, but one might forget this. Under the assumption that we want to show as little information as possible about our internal url structures (debatable as an open source plattform), I added a customized version of the `staff_member_required` decorator, which has this behavior built-in, but also customizable. Let me know what you think!

Baschdl: Is there a possibility to override `staff_member_required` with our version project-wide?

Baschdl (changes requested):
http://localhost:8000/use_statistics/view/ shows the 404 page, http://localhost:8000/use_statistics/view redirects to login

bjrne:
The 404 is not intended by behaviour added in this PR, it is a result of the problem described in #157. The login redirect is a result of the login decorator sitting before the other one. We need to discuss what behaviour we want, e.g. what page do you see for a URL that you have access to, but are not logged in yet. This is a decision between obfuscation and good UX.

backend/apps/matching/views/staff_approvals.py
bjrne: We could give it the same name and import it from our repo, but I thought that might be too easy to miss in a future PR (to check whether the import is right when the decorator is used later in the file). I don't think we can override an import from a django package.

bjrne: We could in theory add a test for it though - but that might be overkill.

backend/apps/matching/views/staff_approvals.py
```diff
@@ -53,18 +62,27 @@ def get(self, request, p_type):
 
     def post(self, request, *args, **kwargs):
         post_params = self.request.POST
+        p_type = self.kwargs["p_type"]
         if "delete" in post_params:
-            p = get_object_or_404(Participant[self.kwargs["p_type"]], uuid=post_params["uuid"])
+            if not request.user.has_perm("matching.delete_participant%s" % p_type.lower()):
```
Baschdl: I would also test for `matching.delete_participant%s` and `matching.can_approve_type_%s` with a class-based decorator. A person who can't change the values, should probably also not see these pages.

Baschdl (changes requested):

backend/apps/matching/views/staff_approvals.py
bjrne: I don't understand what you mean - those get tested in lines 27-30. These here are for the actual action. 

I think using a class decorator would be less expressive as it would probably have a long name. If you use them separately as decorators, both would be required. But I think not every approver would have the rights to delete participants and vice versa?

Baschdl: You're right, I missed that. I still think it would be more consistent to have a method decorator and not have decorators on classes, decorators on methods and checks in a method (which makes it hard to check with a quick glance if the right permissions are set). But I see that this decorator would also not be totally self-explanatory.

backend/apps/matching/views/staff_approvals.py
```diff
@@ -64,25 +60,29 @@ def post(self, request, *args, **kwargs):
...
-            p.change_approval(self.request.user)
-            name = p.user
-            text = format_lazy(
-                _("You changed the approval of the participant with email '{name}'."), name=name
-            )
-            messages.add_message(self.request, messages.INFO, text)
+            return self.post_approve(request, p_type, args, kwargs)
+        return self.get(request, p_type)
+
+    def post_delete(self, request, p_type, *args, **kwargs):
```
Baschdl, line 76: Didn't we want to add the decorators here?

backend/apps/matching/admin.py
```diff
@@ -23,6 +23,42 @@ def participant_required(function=None, redirect_field_name=REDIRECT_FIELD_NAME,
...
 
 
+def can_approve_type_in_url(function=None):
+    """
+    Check that the user at hand can approve objects of given p_type
+    """
+
+    def actual_decorator(function):
+        def new_func(request, p_type, *args, **kwargs):
+            if request.user.has_perm(p_type):
```
bjrne: `p_type` here is either wrong or very misleading I think. It should be the name of the permission.

backend/apps/matching/admin.py
```diff
@@ -23,6 +23,42 @@ def participant_required(function=None, redirect_field_name=REDIRECT_FIELD_NAME,
...
+def can_delete_type_in_url(function=None):
+    """
+    Check that the user at hand can approve objects of given p_type
+    """
+
+    def actual_decorator(function):
+        def new_func(request, p_type, *args, **kwargs):
+            if request.user.has_perm(p_type):
+                return function(request, p_type, *args, **kwargs)
+            raise Http404
```
bjrne: The user is already authenticated, shouldn't this raise or return a 403 page? As we decided in favor of using the right status codes.

backend/apps/matching/admin.py
bjrne: Probably something like
```suggestion
            if request.user.has_perm("can_aporive_type_" + p_type.lower()):
```
?

backend/apps/matching/admin.py
```diff
@@ -23,6 +23,42 @@ def participant_required(function=None, redirect_field_name=REDIRECT_FIELD_NAME,
...
+
+        return new_func
+
+    if function:
+        return actual_decorator(function)
+    return actual_decorator
+
+
+def can_delete_type_in_url(function=None):
```
bjrne: I'm not sure anymore whether decorators with this much logic make sense, it depends on the type of user and a permission dynamically.
And as I, a member of this team and very familiar with this specific piece of code, took almost a minute to figure out what `in_url` means, I am unsure about it even more... I first thought it was another permission stating where you are allowed to delete users.

bjrne:
This will also wait for a decision on permission names (refer to #183).

bjrne:
Will test thoroughly again and also revise checks for buttons on staff-profile.

PR #144, opened 2020-07-11
Use BASE_URL instead of hardcoded url for facebook open graph

### LINKED ISSUE #127: m4h url is hardcoded in some metas
match4healthcare.de is hardcoded in some meta tags in `base.html`. We want to have this dynamically (there was maybe a problem) or at least editable
https://github.com/match4everyone/match4everything/blob/e9578f184bab4720422d7e48292e5f48883ba01d/backend/templates/base.html#L34-L37

Baschdl:
I introduced `django-settings-export` to easily access config variables in the templates. What do you think about this?

Fixes #127


bjrne (approved):
Works :+1: 
I will not merge this yet so others can have a look at the plugin if they care.

PR #146, opened 2020-07-11
Implement handlers right, add styled 403 page

### LINKED ISSUE #145: Implement 404 Handlers right
The [django docs](https://docs.djangoproject.com/en/3.0/ref/urls/) mention that custom handlers should return the correct HttpResponse, for example `HttpResponseNotFound.` for a 404. Currently, we use the standard `HttpResponse`.

bjrne:
Closes #145

While implementing this, I noticed through manually setting the status codes we already had had the correct behavior, but using the right classes is even better.

Also adds a styled 403 header.


backend/locale/de/LC_MESSAGES/django.po
```diff
@@ -431,37 +431,35 @@ msgstr "Englisch"
...
-msgstr ""
-"\n"
-"        404 Error\n"
-"    "
+msgid "403 Error"
+msgstr "403 Error"
 
-msgid "This page could not be found."
-msgstr "Diese Seite konnte nicht gefunden werden."
+msgid "You don't have the permissions to access this page."
```
Baschdl, line 458: To be picky, if we want to be consistent, the page should be in German. We could also open an issue to switch German -> English, to be easier usable in an international context

Baschdl (approved):
Looks good but couldn't test it yet

backend/templates/403.html
```diff
@@ -0,0 +1,74 @@
...
+{% load i18n %}
+
+{% block title %}
+    {% blocktrans trimmed %}
+        403 Error
+    {% endblocktrans %}
+{% endblock %}
+
+{% block header %}
+    <script type="text/javascript">
```
maltezacharias: Remove, not needed

backend/templates/404.html
```diff
@@ -3,7 +3,7 @@
 {% load i18n %}
 
 {% block title %}
-    {% blocktrans %}
+    {% blocktrans trimmed %}
```
maltezacharias, line 6: neat, need more of those!

maltezacharias (changes requested):

backend/locale/de/LC_MESSAGES/django.po
Baschdl: We discussed that we want to slowly transition to purely english source pages, so that's fine.

backend/templates/403.html
maltezacharias: I meant the whole block. There is no e-mail on that page, I'll commit it shortly

maltezacharias (approved):
I tested it locally. When clicking on the button I get returned to django admin login form, but I suppose that's because I have no pages loaded yet. It sends you to `/` which should be right.

Also in `backend/match4everyone/urls.py` there are variables handler404 and handler500 that I suspect can be removed.

backend/templates/403.html
Baschdl: @maltezacharias You can select multiple lines for a comment if you click and drag

PR #149, opened 2020-07-12
Remove p_type from urls that do not need it

### LINKED ISSUE #148: Remove p_type from deletion routes and activation routes
Depends on #105  and #126

bjrne:
Closes #148.

### DIFF
--- a/backend/apps/matching/templates/messages/deleted_user_ask.html
+++ b/backend/apps/matching/templates/messages/deleted_user_ask.html
@@ -46,9 +46,9 @@ Are you sure you want to delete your account?
   <div class="card login-card text-center align-items-center">
       <h4>{% blocktrans %}Are you sure you want to delete your account?{% endblocktrans %}</h4>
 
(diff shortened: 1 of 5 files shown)


Baschdl:
Do they really don't need it? Did you think about use cases where one wants to display different pages? Maybe to note that one deletion has more effects than another (e.g. "No candidates are able to contact you", "You will not be able to communicate with any members of our platform")

feeds:
I think in the cases you mentioned @Baschdl  one can still retrieve the information about the participant from the active user. So since no other user but yourself should be able to 'delete_me' its generally fine.

Only issue could be editing with django cms. But I would go for it as @bjrne proposed.

feeds:
I think in the cases you mentioned @Baschdl  one can still retrieve the information about the participant from the active user. So since no other user but yourself should be able to 'delete_me' its generally fine.

Only issue could be editing with django cms. But I would go for it as @bjrne proposed.

Baschdl:
We don't intend right now to make pages other than home, about, imprint etc. editable

feeds:
then this change is just what we want

Baschdl (approved):
Works and looks fine

Issue #130, opened 2020-07-10
Add integration tests


bjrne:
- [ ] As a user_a, I want to successfully complete the login page, visit my profile, edit my profile and log out.
- [ ] As a user_a, I want to successfully log in to the website, change my password and successfully log out and in again.
- [ ] As a user_a, I want to successfully reset my password without knowing it.
- [ ] As a user_b, I want to successfully complete the login page, search for user_a's via the map and contact 2 user_a's matching some filter with a zipcode that is not mine and a 1000km radius.
- [ ] As a staff user, I want to successfully approve type a and b users.
- [ ] As a staff user, I want to view access and usage statistics.
- [ ] As a staff user, I want to change a text label in the cms and see the updated page online.
- [ ] As an intermediate staff user, I want to give other staff users permissions for approving users.
- [ ] As an interested person without an account, I want to be able to open all static pages in order to inform myself. (about, imprint, home, data agreement, possibly press)

Feel free to add more ideas :)

PR #152, opened 2020-07-13
Make approvals searchable

### LINKED ISSUE #151: Make approvals searchable by mail
As a staff member with approval rights, I want to search for specific users to approve/unapprove

Discussed in #68

Baschdl:
Fixes #151


backend/apps/matching/templates/staff/approve_participants.html
```diff
@@ -8,6 +8,7 @@
 
 {% block header %}
 <link rel="stylesheet" href="{% static 'css/form.css' %}">
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
```
Baschdl: @maltezacharias Is this still the way do it?

backend/apps/matching/urls.py
```diff
@@ -56,7 +56,7 @@
     ####################################
     path("profile_staff", views.StaffProfileView.as_view(), name="staff_profile"),
     path(
-        "staff/<p:p_type>/approve",
+        "staff/<p:p_type>/approve/",
```
Baschdl: It strangely stopped working without the `/`

bjrne: One more reason for #130! :tada: :wink:

backend/apps/matching/templates/staff/approve_participants.html
```diff
@@ -39,15 +40,25 @@ <h2>
             {% blocktrans %}These participants require approval.{% endblocktrans %}
         </div>
         <div class="card-body">
+            <form>
```
Baschdl, line 44: `csfr_token` is still missing

backend/apps/matching/urls.py
bjrne: And for #157 now :grin:

backend/apps/matching/templates/staff/approve_participants.html
Baschdl: done

backend/apps/matching/templates/staff/approve_participants.html
maltezacharias: We haven't used CDNs because of the privacy issues throughout the project. As for the CSS, depends. If there is no JS on the page (so far there isn't) yes, that's a possible way. If there is also JS code I tend to load the CSS in there (e.g. `frontend/src/main.js`)

feeds (approved):
Approve, given that you move the files to the correct folder @Baschdl, just like malte mentioned above

maltezacharias:
Font-Awesome is now loaded via the webpack, to test you need to (re)build the Javascript Bundles. With docker this happens automatically, when using runserver please use `docker-compose run --rm frontend npm run build` (see Readme)

feeds:
But then it is ready for merge? :)

PR #159, opened 2020-07-13
Add sites to change mail limit for individual participants

### LINKED ISSUE #68: Enable staff to change the max number of contact requests a person can send per day and indroduce the field at all
We did not have this ability in m4h and this was slightly annoying for the admins

The field needs to be introduced on the abstract model class of participantA/B

@Baschdl:
Use the same functionality and design like the approval pages but to provide the possibility to increase the mail limit

Baschdl:
Uses search from #152 and permission check on profile page like #132

@bjrne What is the workflow to create the `can_change_mail_limit_*` permissions?

Partly improves #68, but still needs to use this field in the email sending etc.


bjrne:
Depends on #70 and that depends on #124.

feeds (approved):
I fixed the migrations and the link to the participants (which is currently throwing 404 since participants can only fix themselves, but this will be fixed in #120)

PR #161, opened 2020-07-13
Use another url for /admin/

### LINKED ISSUE #158: Hide /admin/ if a user is not logged in
Redirect `/admin/` to 404 if the user is not logged in to make it a bit harder for automatic scripts to find our login form

Baschdl:
Use another url for `/admin/` to make it a bit harder for script kiddies

Fixes #158


bjrne (dismissed):
Works, but does it really matter? Our login is at `/login/`, and I can log in there with my admin credentials, so I don't really see the point. One could of course also obfuscate the login link too, and only let people login through there and only let logged-in users visit the admin page. But I suspect this adds more waste than it add security...
I think this needs to be thought through thoroughly (:wink:) to be one element in a coherent security structure and does not really add value as is.

Baschdl:
Not having changed `/login/` yet isn't really an argument for me. We can decide together that it doesn't matter and just close this PR but the login sites of the popular frameworks are targeted in my experience.

feeds:
the login is actually under `matching/login` so that should be safe

feeds (approved):

maltezacharias:
I don't see the need for this, enforcing and encouraging safe passwords seems better than changing the default URLs in my mind. It is still only security through obscurity. It gives a false sense of security in my opinion. But If you need to have it feel free to merge.

Baschdl:
@maltezacharias We basically have two really safe options: preventing passwords which are too easy for a dictionary attack or use something like fail2ban. In my opinion, none of them will be implemented in the remaining time, so that's the best we can do.

Issue #147, opened 2020-07-11
Replace all hardcoded urls with url tags in templates


bjrne:

Baschdl:
Do you find something besides facebook, twitter, youtube and github?

bjrne:
- For example 'matching/password_change' in the participant dashboard, I am sure that I have seen more.

Baschdl:
Ah, you're referring to the question @maltezacharias already asked in https://github.com/match4everyone/match4everything/pull/105#discussion_r452454882

PR #163, opened 2020-07-14
Add trailing slashes to urls and redirect with names

### LINKED ISSUE #157: Change all routes to end in slashes
After a quick research, I think the best practice is for links to have a trailing slash, and to redirect links without one to that url (which is built-in behaviour of django, like documented [here](https://docs.djangoproject.com/en/dev/ref/settings/#std:setting-APPEND_SLASH)). 

Currently, routes whose paths do __not__ end in slashes but are entered with one, result in a 404.

There is more information on the [to-slash-or-not-to-slash](https://webmasters.googleblog.com/2010/04/to-slash-or-not-to-slash.html) from Google from 2010.

Do you have any take on this @maltezacharias?

bjrne:
Closes #157


Baschdl (changes requested):
You may also want to add a slash to `use_statistics/view`. The B button on the start page leads to a 404 but I think that the `href` is incorrect: `matching/signup_b/` (Done in PR for #147). logout with `matching/profile_staff/logout/` also doesn't work.

bjrne:
Signup for B works for me. The logout url from the staff profile page is `/matching/logout/` for me and also works. Maybe there was a bug with `staff_profile` and `profile_staff`, but for me it works now.

Baschdl (approved):
Works now

PR #160, opened 2020-07-13
Make admin page filterable and searchable

Baschdl:
Closes #154 and closes #155

Issue #51, opened 2020-06-21
Create a hierachy of technical and regular influencer admins


feeds:
The `createsuperuser` command receives all permissions by default and can log into the django admin user interface. 
We do not want to grant this permission to all '_matching_supervisors_'.

Thus, we need to set up the correct groups and permissions at all places.
From django: 
- `is_staff`has access to the django admin interface
- `is_superuser` has **all** permissions

Suggestion:
- add `is_matching_supervisor` 
- matching staff interface is accessible to all of the three above groups and is_staff users see a link to the admin interface
- create a facitlity (django command) to create a matching supervisor with a dialog that asks wether it should have the permissions a supervisor can have (approve_a, approve_b, view_access_stats, view_usage_stats...) 
- remove the Staff class (we do not really need it)

Otherwise we could also only use "is_staff" and allow access to the django admin panel to everyone

feeds:
@bjrne this is what we talked about (also in #22, but I think you missed that staff users can already access the admin panel)
@Baschdl what kind of setup do you need for the cms? Would this interefere somehow?

Baschdl:
We would introduce another group "page editors" and maybe "(news) blog editors" and "press editors". This would be independent of all the other groups.

feeds:
So many people :D

Baschdl:
@match4everyone/match4everyone How should a user, which we called `staff` before, sign up? Do we have an additional hidden signup page or does a `superuser` or `is_matching_supervisor` create them?

maltezacharias:
> @match4everyone/match4everyone How should a user which we called `staff` before sign up? Do we have an additional hidden signup page or does a `superuser` or `is_matching_supervisor` create them?

I don't understand the question, I think you left out a word or so`

Baschdl:
@maltezacharias Added two `,`

maltezacharias:
Ahh thanks. I would suggest: Using the built-in django admin a superuser can adjust the roles of staff users. We would need to modify the Django user admin for this, by default you can't search for users (which is a problem with a userbase of >1000) and you can't create new users properly because the password input field expects the hashed password. 

I think both should be easily doable by extending the default useradmin. It would also be beneficial because you can easily edit user groups there for all kinds of permission related issues.

maltezacharias:
Here's a link to the pertinent django docs for search fields: https://docs.djangoproject.com/en/3.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.search_fields

[...]

PR #165, opened 2020-07-14
Configure admin interface to add users

Baschdl:
It is now possible to add new users on the admin interface without calculating the hashed password by hand

Discussed in #51

Depends on #160


josauder (approved):
LGTM, tested locally and it works

PR #169, opened 2020-07-14
Show a warning in the dashboard if the user is deactivated

### LINKED ISSUE #168: Show a warning in the dashboard if the user is deactivated

Baschdl:
A message every time a user visits the dashboard

Closes #168


feeds (approved):
If the translations are fixed, this works like a charm

PR #173, opened 2020-07-14
0 - Create persistent filters with dashboard and contact possiblities

### LINKED ISSUE #37: Create a persistent filter that can be applied to a participant set

### LINKED ISSUE #176: Optional name for filters
As a user with many filters I can easily be lost with all my filters but as a users with only few filters I don't want to understand what a name of a filter is.

Add a standard name or empty text field to the list of filters which can be assigned by power users but doesn't have to by standard user

feeds:
You can use the persistent filters from a participants dashboard

Fixes #176
Closes #37


feeds:
I agreed with @maltezacharias that this should be merged as it is a useful mockup that will obtain nicer UI elements in the future, so it is useful to merge this into staging

backend/locale/de/LC_MESSAGES/django.po
```diff
@@ -8,22 +8,46 @@ msgstr ""
...
 "Language: de\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
-"X-Generator: Poedit 2.3\n"
+"X-Generator: Poedit 2.3.1\n"
+
+msgid "is exactly"
+msgstr ""
```
bjrne:
```suggestion
msgstr "ist genau"
```

Baschdl:
I don't understand all these filter PRs. @feeds Is this description still up-to-date?

> Create a new Filter that is saved in http://127.0.0.1:8000/matching/A/filter/create
> or call the rest api to filter the existing participants with get parameters http://127.0.0.1:8000/matching/api/A/info/list (see OPTIONs-button)

backend/apps/matching/urls.py
```diff
@@ -44,8 +44,23 @@
     path("map", views.map_view, name="map"),
     path("<p:p_type>/participant_JSON", views.map_JSON, name="participant_JSON"),
     ####################################
-    #  List view
+    #  List view / filtering
     ####################################
+    path(
+        "api/<p:p_type>/info/list",
```
Baschdl: An endpoint where you can get all of our data sounds like a bad idea

backend/apps/matching/views/participant_info_filter.py
```diff
@@ -0,0 +1,30 @@
...
+from django.urls import reverse
+from django.utils.decorators import method_decorator
+from django.views.generic.edit import CreateView
+
+from apps.matching.forms import ParticipantFilterForm
+
+logger = logging.getLogger(__name__)
+
+
+@method_decorator([login_required], name="dispatch")
```
Baschdl, line 15: Please check for the right permissions

Baschdl (changes requested):
In m4h an attacker had to hop through Germany by repeatedly changing their location by 50km (?) to copy all of our data, now it's just an API call 😬 Let's discuss @match4everyone/match4everyone

feeds:
Sorry, the added user functions are

- go to dashboard as a participant you have two new options: search for opposite participants and view past searches

- search: filter possiblity on top that will be replaced by a js filter layout, table that shows results below that will also be replaced by a js component that makes calls to an api route that already exists in this PR `api/A/info/list`. One can then select several participants with a checkbox. Then a popup opens to promt subject and message. On send the filter is saved and the contacted people are put in a database. Return to dashboard. 

- saved search: list of searches that were already made. Shows how many matches exists in the database live. _later_ should show how many of those were contacted, button for search again, Button for contact all i did not contact before. 

db changes :
- dynamically generates the filters from the config. i changed the properties to define what filters go with each field.

Baschdl:
- [ ] Is it on purpose that only search fields for `name`, `contact` and `phone number` exist when http://127.0.0.1:8000/matching/A/filter/create/ also allows me to search for boolean etc. fields?
- [ ] No possibility to name the filter but http://localhost:8000/matching/B/filter/list/ includes `name`
- [ ] No description: 
image
- [ ] I can search for letters in the phone number

backend/apps/matching/urls.py
maltezacharias: Same as we had before. We need to be able to retrieve the results.
As Freya sad above this is a workable mockup and since we are still working on a release ready product there is no need that every PR needs to be feature complete. The idea was that it should be easier to review this way. If we include all follow-up changes in this PR it will be enormous, which was what you criticized with my big PRs.

Next step will also be to add an obligatory location and distance filter to this api call. But we can never prevent scraping of all publicly accessible data. It might be a good idea to implement a rate limiting of some sort to make this harder. Or an enforced delay between queries, this would be a new issue

maltezacharias:
> * [ ]  Is it on purpose that only search fields for `name`, `contact` and `phone number` exist when http://127.0.0.1:8000/matching/A/filter/create/ also allows me to search for boolean etc. fields?
> * [ ]  No possibility to name the filter but http://localhost:8000/matching/B/filter/list/ includes `name`
> * [ ]  No description:
>   image
> * [ ]  I can search for letters in the phone number

The UI is only a mockup, we are working on an endpoint that provides the available filters to the Javascript side and an UI that dynamically builds on this

backend/apps/matching/urls.py
Baschdl: When is 

> before

for you?

> In m4h an attacker had to hop through Germany by repeatedly changing their location by 50km (?) to copy all of our data, now it's just an API call 😬

Baschdl:
@maltezacharias Then please migrate the points which you are working on to the corresponding issue. 
It's hard to review without knowing what the intended functionality is and, even more important, not knowing which functionality is not intended for this PR.

backend/apps/matching/urls.py
Baschdl: This point is really not about having a release ready product but about not introducing an even bigger vulnerability

Baschdl: Discussed in #177

Baschdl:
@feeds and me opened #176 and #177

feeds:
Now also fixes #176

Allows to
- modify search
- contact everyone that was contacted before
- see new matches

feeds:
I still need some input on what people should actually see in the filter - right now its not pretty
image
Also we would need to figure out the logic behind what sort of matches new ones are. @Baschdl

backend/apps/matching/views/participant_filter_api.py
```diff
@@ -0,0 +1,50 @@
...
+        filter_specification = ParticipantInfoFilterSet[p_type].filter_spec()
+
+        return {
+            "name": view.get_view_name(),
+            "description": view.get_view_description(),
+            "actions": {"GET": filter_specification},
+        }
+
+
+# how do we want to manage access here? Only let users be searchable for each other?
```
Baschdl, line 32: @feeds Would it be a different logic than in #175?

feeds:
Also fixes #176

feeds:
oops

PR #174, opened 2020-07-14
Route for filter type options

feeds:
Depends on #173 

Introduce a route for json filter options:

`http://127.0.0.1:8000/matching/api/A/info/filter-options/`


Baschdl (changes requested):
Could you please add a documentation about when and how to use it as the main audience are presumably developers?

feeds:
Included in the filter-ecosystem PR #196

PR #175, opened 2020-07-15
Add configuration of visibility of profile page

### LINKED ISSUE #172: Visibility of pages should be specifiable in configuration
@feeds: Depends on #120 

- Can an A see the profile of all other As or only their own?
- Can a B see the profile of all other Bs or only their own?
- Can an A see the profile of all other Bs?
- Can a B see the profile of all other As?

Baschdl:
Reads configurations and sets visibility accordingly.

Closes #172


feeds:
how about `@required_at_least_one(['view_a',view_xy','was_nice_to_santa'])` ?

backend/apps/matching/admin.py
```diff
@@ -61,3 +61,46 @@ def logged_in_not_permitted(function=None, login_url="/"):
...
+            if p_type == "A":
+                for restriction in restrictions_A:
+                    passes_test += restriction(u)
+            elif p_type == "B":
+                for restriction in restrictions_B:
+                    passes_test += restriction(u)
+            passes_test += profile_own(uuid_accessed=uuid, uuid_own=u.participant().info.uuid)
+            if passes_test > 0:
+                return function(request, p_type, uuid, *args, **kwargs)
+            raise Http404
```
Baschdl: @bjrne Did you also implement something like this for 403 or what's the way to do it?

bjrne: Well a 403 page is implemented now, but you would need to link to it via its name "403". Http404 is the only Exception implemented by django in this way unfortunately:

https://github.com/django/django/blob/3d16496037fbb8a6bbc6b6e354fa4f5eb65e6cea/django/http/response.py#L539

feeds:
Closed in favour of #200

PR #181, opened 2020-07-18
Implement API rate limits

### LINKED ISSUE #177: Limit API requests
As an ill intented user I have access to all of our data via the filter api.

Add a limit for the number of results a api request returns and limit the total number of api requests.

Could also include a daily and total limit and a minimal time between requests.

Baschdl:
With this settings, a user is only allowed to retrieve 15*100=1500 users per day. @match4everyone/match4everyone Currently, this has to be changed in the `settings/common.py` and not like in `A.py` and `B.py`. Do we want to expose such general settings also in something like a `general.py` or should we just document that the user has to change that in `settings/common.py`?

Closes #177

Depends on #173

### DIFF
--- a/backend/apps/matching/utils/rest_framework/pagination.py
+++ b/backend/apps/matching/utils/rest_framework/pagination.py
@@ -1,0 +1,5 @@
+from rest_framework.pagination import LimitOffsetPagination
+
+
+class StrictLimitOffsetPagination(LimitOffsetPagination):
+    max_limit = 15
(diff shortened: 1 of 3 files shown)


feeds:
I would raise the limit per minute. Skipping through more than 5 pages and adapting the filter several times (= one requests each) is quite normal. I would set it to 40 the least. Bigger crawlers will be stopped by the hour wise rate limit.

feeds:
I think its a rather technical setting, so it should go to `settings.py`

Baschdl:
After long negotiations, we settled for 15/min

Baschdl:
@feeds Mind leaving a review? 😄

Baschdl:
@feeds @bjrne @maltezacharias Anyone?

PR #182, opened 2020-07-18
Tidy up scripts

Baschdl:
@bjrne @maltezacharias What do you think? Do you have an idea for a cleaner solution?


scripts/delete_db_and_setup.sh
```diff
@@ -1,6 +1,10 @@
-rm backend/run/db.sqlite3
-python3 backend/manage.py migrate
-echo "Creating account with username \"admin\". "
-echo "Please enter your desired email and password:"
-python3 backend/manage.py createsuperuser  --username admin
-python3 backend/manage.py loaddata backend/cms_export.json
+#!/bin/bash
```
maltezacharias:
```suggestion
#!/bin/bash -e
```
(Will exit when a command fails)

scripts/delete_db_and_setup.sh
```diff
@@ -1,6 +1,10 @@
...
+# To run script in Docker call for example "delete_db_and_setup.sh 'docker-compose exec backend'"
+EXEC_ENV="$*"
+
```
maltezacharias: It would be nice if you could run it from anywhere, but I don't know how to do that..
I tried this:
```suggestion
DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )" && cd $DIR/../
```

but then it would be a bit unexpected that you need to provide a docker-compose command taking into account the directory switch. On the other hand you wouldn't need to care about where you are when you call that script, it would always be `docker-compose -f docker-compose.yml -f docker-compose.prod.yml`

What do you think @Baschdl ?

scripts/delete_db_and_setup.sh
```diff
@@ -1,6 +1,10 @@
...
+$EXEC_ENV python3 manage.py reset_db
```
maltezacharias: Neat!

maltezacharias (approved):

maltezacharias:
Please review above suggestions. The PR looks fine to me with or without them.

scripts/delete_db_and_setup.sh
Baschdl: I think we would also have to specify the path of `backend/cms_export.json` dynamically. Maybe that's a bit too much engineering?

maltezacharias: Probably.. The suggestion I wrote switches to the project root dir reliably no matter from where you call the script (copied from StackOverflow, there is an error condition somewhere with symlinks that I forgot, but as you said... Didn't want to over-engineer. I wouldn't invest more time I think

bjrne (commented):
Is the change in permissions intended?
100644 → 100755

bjrne:
> Do you have an idea for a cleaner solution?

I don't :sweat_smile:

Baschdl:
> Is the change in permissions intended?

Yes, it makes it executable

scripts/delete_db_and_setup.sh
Baschdl: Not sure what to do here, I really like the idea but it may be a bit of an unexpected behavior for someone using it. We should probably maximize for the understandability for a new user

backend/apps/matching/management/commands/createfakeusers.py
```diff
@@ -17,7 +17,7 @@ class Command(BaseCommand):
     # has to be "help" because we inherit from django manage.py Command, thus ignore A003
     help = (  # noqa
         "Populates the database with fake users or deletes them."
-        "Every user has their email as their password."
+        "Every user has type (A|B) as their password."
```
Baschdl: @bjrne But why? I find it a lot more straightforward to just copy down the mail address.

backend/apps/matching/models/user.py
```diff
@@ -34,7 +35,7 @@ def new(email, pwd=None, **kwargs):
         if pwd is not None:
             user.set_password(pwd)
         else:
-            user.set_password(email)
+            user.set_password(get_random_string(length=32))
```
Baschdl: I think we should switch to `def new(email, pwd, **kwargs)` and don't have any process of setting an automatic password.

scripts/clean_db_and_setup_with_fake_users.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -e
+# To run script in Docker call for example "delete_db_and_setup.sh 'docker-compose exec backend'"
+EXEC_ENV="$*"
+DJANGO_SUPERUSER_PASSWORD="admin"
+
+$EXEC_ENV echo "DO NOT USE THIS SCRIPT IN PRODUCTION"
```
Baschdl: Why don't we leave it general enough for production and development? I have actively decided against including the `createfakeusers` etc.

bjrne: Well then just delete all my commits. I have an alias for this, but then you could just make the script docker only. If you need to enter this many things and also create fake users manually, I won't use it for development, and I suspect other wont aswell.

backend/apps/matching/models/user.py
bjrne: Sure

backend/apps/matching/management/commands/createfakeusers.py
bjrne: Copy is two clicks plus keystroke, this is one click plus keystroke ;) I you don't like it reverse it, I have several logins saved as credentials in the browser.

Baschdl: It's "Helper" or "Institution" now, so the mail is easier

scripts/clean_db_and_setup_with_fake_users.sh
Baschdl: Merged both scripts

scripts/delete_db_and_setup.sh
bjrne: Right now it needs to be called with `bash ../scripts/delete_and_setup_db.py` right?

Baschdl: Yes, `cd backend && bash ../scripts/delete_and_setup_db.py` if you don't use docker

scripts/delete_db_and_setup.sh
```diff
@@ -1,6 +1,29 @@
...
+#!/bin/bash -e
+# To run script in Docker call for example "delete_db_and_setup.sh 'docker-compose exec backend'"
+#MODE={$1:-"None"}
+MODE=$1
+if [[  $MODE != 'DEV' && $MODE != 'PROD' ]]; then
+  echo "Supplied mode is not supported. We only support 'DEV' and 'PROD'"
+  exit 1
+fi
+if [ $MODE == 'DEV' ]; then
+  echo "DO NOT USE THIS SCRIPT IN PRODUCTION"
```
bjrne: Maybe this hint is misleading when there is a production version in the same script? But I don't have a better idea.

bjrne (approved):
This needs some documentation, but then it's ready.

scripts/delete_db_and_setup.sh
Baschdl:
```suggestion
  echo "DO NOT USE THE 'DEV' MODE IN PRODUCTION"
```
@bjrne How about this?

bjrne: Sounds funny, but better than everything I've come up with.

scripts/delete_db_and_setup.sh
maltezacharias: With the proposed change it should work from everywhere.

Baschdl: > would be a bit unexpected that you need to provide a docker-compose command taking into account the directory switch

@maltezacharias I'm not sure what you mean with this. It's always relative to `backend` if you use docker, right?

scripts/delete_db_and_setup.sh
```diff
@@ -21,7 +21,7 @@ if [ $MODE == 'PROD' ]; then
   echo "Please enter your desired email and password."
   $EXEC_ENV python3 manage.py createsuperuser --username admin
 else
-  $EXEC_ENV python3 manage.py createsuperuser --username admin --email admin@examplee.com --no-input
```
bjrne: This was actually intentional in order to preserve the superuser when deleting all fakeusers with the script. But I can live with it both ways since I rarely use the fakeuser-deletion.

bjrne (approved):
Let's merge this? @Baschdl

Baschdl:
@bjrne I would wait for the answer from @maltezacharias but I'm also missing this script in the other branches

scripts/delete_db_and_setup.sh
Baschdl: We can add it again. Without documentation this looks like a typo 😬

scripts/delete_db_and_setup.sh
maltezacharias: My suggested line always checks where the script is located (the file itself) and then changes the workdir to the directory above. So you would always execute your commands from the repo root. That would mean you can either call it in one of the following ways:

```
# Using an absolute path from anywhere
/home/user/match4everything/scripts/delete_db_and_setup.sh  # for native django
/home/user/match4everything/scripts/delete_db_and_setup.sh docker-compose exec backend # for dev docker
/home/user/match4everything/scripts/delete_db_and_setup.sh docker-compose -f docker-compose.yml -f docker-compose.prod.yml exec backend # for dev docker

# From scripts dir
cd scripts
./delete_db_and_setup.sh  # for native django
./delete_db_and_setup.sh docker-compose exec backend # for dev docker
./delete_db_and_setup.sh docker-compose -f docker-compose.yml -f docker-compose.prod.yml exec backend # for dev docker

# from repo root
./scripts/delete_db_and_setup.sh  # for native django
./scripts/delete_db_and_setup.sh docker-compose exec backend # for dev docker
./scripts/delete_db_and_setup.sh docker-compose -f docker-compose.yml -f docker-compose.prod.yml exec backend # for dev docker
```

All of these should work equally (If I didn't make a typo)

maltezacharias:
> @bjrne I would wait for the answer from @maltezacharias but I'm also missing this script in the other branches

please ping on telegram if the answer is not sufficient, don't want to keep you waiting

PR #183, opened 2020-07-18
Use url_name in urls

### LINKED ISSUE #113: Clean up links

Baschdl:
Use the `url_name` from the configuration as `p_type` in the urls etc.

"A" and "B" are still in the dictionaries for programmer's convenience 

Closes #113


Baschdl:
@feeds Can we use instead of https://github.com/match4everyone/match4everything/blob/2341e57c775cfbb773710d3d17d7f753a5403e08/backend/apps/matching/models/participant.py#L63-L66
something where we first filter for 'p_type' and then filter for 'is_approved'? Alternatively, we would have to also name the groups according to the names in the config.

But we would have similar problems for `if not approver.has_perm(`

Baschdl:
@bjrne `isort` is blocking this one like #175 but possibly one other formatting issue

backend/apps/matching/models/participant_info_location.py
```diff
@@ -80,4 +82,9 @@ class ParticipantInfoLocationB(AbstractParticipantInfoLocation):
     )
 
 
-ParticipantInfoLocation = {"A": ParticipantInfoLocationA, "B": ParticipantInfoLocationB}
+ParticipantInfoLocation = {
+    "A": ParticipantInfoLocationA,
+    A.url_name: ParticipantInfoLocationA,
+    "B": ParticipantInfoLocationB,
+    B.url_name: ParticipantInfoLocationA,
```
bjrne: This does not look right, correct me if I'm wrong.
```suggestion
    B.url_name: ParticipantInfoLocationB,
```

backend/apps/matching/management/commands/createfakeusers.py
```diff
@@ -29,30 +31,38 @@ def add_arguments(self, parser):
         )
 
         parser.add_argument(
-            "--add-a", nargs=1, help="Add [N] new students to the poll",
+            "--add-{a}".format(a=A.url_name),
```
bjrne, line 34: Lets use
```suggestion
            "--add-{a}".format(a=A.url_name.lower()),
```

bjrne (dismissed):

backend/apps/matching/management/commands/createfakeusers.py
Baschdl: It's `url_name = name.lower()` in `A.py` and `B.py`

bjrne: Ah, now i get it.

feeds:
- [x] change the urlconverter
- [x] urlencode for url_name

feeds (approved):
I added the changes I wanted to see @Baschdl. If you are fine with this please merge :)
And @bjrne you're requests are addressed

PR #184, opened 2020-07-18
Fix sort import problem

bjrne:
Imports from `match4everyone` and `apps.*` are now grouped in one group. Flake8 likes it, and it makes sense to only have one group of first_party imports.


maltezacharias:
I would like to approve, but I don't understand it, can you elaborate a little:
- what this does 
- when you would need to do what changes? 
- Will this modify the import sort order for the existing project files?

bjrne:
Former line 11 lists the import groups we want to have, former line 10 defined what belonged to the separate  group `match4everyone`, since we defined it, so it didn't know. Flake is smart and gets that both `apps.*` and `match4everyone` are local packages and wants to combine them. So the solution is either to put them separately and modify flake8, or put them together and remove this custom logic in the isort config.

Since we never had the case of importing from both the apps and the match4healthcare directory before, this will not change any existing code :grin: (as you can check by running `pre-commit run -a` on this branch)

maltezacharias (approved):

PR #187, opened 2020-07-19
1 -View and control successful matches

### LINKED ISSUE #188: Allow users to see who they contacted and when and share their email via our plattform

feeds:
Allow participants to view contact requests and reply to them, accessible via the dashboard
image
image
image

Closes #188


bjrne:
Seeing it now stacked in github, I would even say the contact histories can be combined or even shown directly in the dashboard, but I like thhe toplevel navigation with the icons too :)

feeds:
Yes, nice to have, but its a bit annoying to combine them with django, so i'll continue and do the required tasks  @bjrne  ;)

PR #191, opened 2020-07-20
Make locations editable

### LINKED ISSUE #190: Allow a user to edit their location

feeds:
So far, there was no possibility to edit your location in your profile, now you can do it via edit info.
This fixes #190


backend/apps/matching/templates/participant/participant_info_edit_form.html
```diff
@@ -14,23 +12,95 @@
...
+                <hr>
+                <input type="submit" class="btn btn-success float-right" value="Save">
+                </form>
             </div>
         </div>
 
     </div>
+
+    <!-- inspired from : https://simpleit.rocks/python/django/dynamic-add-form-with-add-button-in-django-modelformset-template/ -->
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>
```
Baschdl, line 96: Should probably also be included with webpack, @maltezacharias?

Baschdl (dismissed):
Can we show errors at the top of the page instead of under the heading "Locations"?

Works and fine otherwise 😄

feeds:
> Can we show errors at the top of the page instead of under the heading "Locations"?
> 
> Works and fine otherwise smile

I did this intentionally, since with long info-configs you need to scroll a while until you see them.
I would leave it like this if it is ok.

backend/apps/matching/templates/participant/participant_info_edit_form.html
feeds: If @maltezacharias  fixes this it would be good to merge then?

feeds:
image

feeds:
@Baschdl could you please release the change request, we talked about it when I made the last commits ;)

backend/apps/matching/templates/participant/participant_info_edit_form.html
Baschdl: @maltezacharias

PR #200, opened 2020-07-22
Alternate implementation of #175

bjrne:
This is an alternative to #175, which has all the logic moved to user creation, thus only needing two built-in django permissions which can easily be checked for.

Given that the profile_view-view is probably the only time this is used and we have to implement the logic in dispatch for #120, I think not going with decorators here is a fine solution.

### DIFF
--- a/backend/apps/matching/models/user.py
+++ b/backend/apps/matching/models/user.py
@@ -1,9 +1,12 @@
 import logging
 
-from django.contrib.auth.models import AbstractUser
+from django.contrib.auth.models import AbstractUser, Group
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils.translation import gettext_lazy as _
+
+from match4everyone.configuration.A import A
+from match4everyone.configuration.B import B
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


backend/apps/matching/models/user.py
```diff
@@ -35,6 +38,25 @@ def new(email, pwd=None, **kwargs):
...
         else:
             user.set_password(email)
+
+        group_is_a = Group.objects.get(name="is_a")
+        group_is_b = Group.objects.get(name="is_b")
+        group_can_view_a = Group.objects.get(name="can_view_a")
+        group_can_view_b = Group.objects.get(name="can_view_b")
+
+        if kwargs["is_A"]:
+            user.groups.add(group_is_a)
```
bjrne, line 48: The `is_A` and `is_B` groups are not necessary for this PR, but we never fixed this before, so I added it here.

feeds:
Later on we have the table where we record matches. In this cases participants should also be able to view each other. So theres more logic needed then, but I guess thats another PR?

Baschdl:
@feeds A participant can see in this table who else also matched with a hospital? I don't think I understand it

PR #204, opened 2020-07-22
Automatically generate favicon from logo.svg

bjrne:
We now generate all favicon sizes and the matching meta tags automatically during the webpack build process. 

Since the icons are generated on every run, I ignored them in git, since are regenerated easily and there might be bit-level differences on different machines due to varying library version, OSes, ...

Open TODOs:
- maybe import theme color from scss, did not work for me.


.gitignore
```diff
@@ -7,6 +7,11 @@ backend/apps/migrations/*
 !backend/apps/migrations/__init__.py
 !backend/apps/migrations/0002_permission_group_creation.py
 
+# Generated by webpack favicon generator
+*.gz
+
+# Generated on every webpack build
+backend/static/img/favicon/*
```
maltezacharias: should generate in frontend/dist, that folder is .gitignored already

backend/templates/base.html
```diff
@@ -7,29 +7,7 @@
...
-    <meta name="application-name" content="&nbsp;"/>
-    <meta name="msapplication-TileColor" content="#FFFFFF" />
-    <meta name="msapplication-TileImage" content="{% static 'img/favicon/mstile-144x144.png' %}" />
-    <meta name="msapplication-square70x70logo" content="{% static 'img/favicon/mstile-70x70.png' %}" />
-    <meta name="msapplication-square150x150logo" content="{% static 'img/favicon/mstile-150x150.png' %}" />
-    <meta name="msapplication-wide310x150logo" content="{% static 'img/favicon/mstile-310x150.png' %}" />
-    <meta name="msapplication-square310x310logo" content="{% static 'img/favicon/mstile-310x310.png' %}" />
-    <meta name="theme-color" content="#ffffff">
-
+    {% include "favicons.html" %}
```
maltezacharias, line 10: pls try if include ../../frontend/dist/favicons.html works

backend/templates/favicons.html
```diff
@@ -0,0 +1,49 @@
+<link rel="shortcut icon" href="/static/img/favicon/favicon.ico">
```
maltezacharias: pls move file to dist

frontend/webpack.config.js
```diff
@@ -148,5 +149,29 @@ module.exports = {
             indent: '\t',
         }),
         new CompressionPlugin(),
+        new FaviconsWebpackPlugin({
+            logo: '../backend/static/img/logo.svg',
+            cache: true,
+            outputPath: './../../backend/static/img/favicon',
```
maltezacharias: pls check if you need to set outputPath at all or whether it would just use the default one from the webpack config. In any case it should be outputted to the dist folder in frontend. Also please see for the other paths, like in Python paths should be constructed using the os-agnostic methods instead of written directly

frontend/webpack.config.js
```diff
@@ -148,5 +149,29 @@ module.exports = {
...
+            // Prefix path for generated assets in generated html
+            prefix: 'img/favicon',
+            inject: false,
+            favicons: {
+                appName: 'match4everyone',
```
maltezacharias, line 161: I would suggest putting this configuration inside backend maybe as json and load it here? But that is probably an extra issue and depending on the docs as right now configs are split to multiple places anyway

maltezacharias (changes requested):
Looks already very nice. Needs some little adjustments for the file locations. Frontend should generate all needed files into dist. That folder is automatically ignored already and copied into the docker image during the image build. Relative paths between backend and frontend/dist are the same in docker and in local development mode, so that should also fix the build problem

frontend/webpack.config.js
bjrne: Yes, In the end, one config json would be a good thing, but also somewhat complicated to implement. I would treat it as a separate issue.

backend/templates/base.html
bjrne: It does not, templates do not support relative links. They do however support multiple locations where to search files, so I added the dist folder to that list in common.py.

frontend/webpack.config.js
bjrne: Done I think? if it is only a folder name without any slashes it should work fine, right?

bjrne:
Please review again @maltezacharias, I hope I have adressed all points (at least travis runs through now).

Please don't merge yet, I will add documentation for this once we decide it is the right way to do it.

When duplicating the logo, I was wondering why we still have all out image files in the backend folder - wouldn't it make more sense to put them into the frontend, now that we have it? And then have a build rule that copies them into the dist folder. This would
- reduce duplication (logos)
- work locally and in docker
- make more sense, as logos are frontend related stuff
- open the door for automatically generating gzipped versions to save bandwith

What do you think?

.pre-commit-config.yaml
```diff
@@ -8,12 +8,12 @@ repos:
...
         exclude: |
                 (?x)(
-                    ^00_old_m4h_matching_code/|
                     ^backend/static/|
                     ^backend/backups/|
                     ^backend/run/|
                     ^database/backups/|
-                    ^database/data/
+                    ^database/data/|
+                    ^backend/templates/favicons.html
```
bjrne: Ooops, still needs to be removed

backend/match4everyone/settings/common.py
```diff
@@ -98,6 +103,7 @@
         },
     },
 ]
+print(TEMPLATES)
```
maltezacharias: Should that be here? :)

maltezacharias (approved):
Looks very good to me, unfortunately I was unable to test.
See the one print statement, I think that ought to be removed. Otherwise well done!

backend/match4everyone/settings/common.py
bjrne: You probably had an old version in your cache, this was already removed in 58fad7d :+1:

.pre-commit-config.yaml
bjrne:
```suggestion
                    ^database/data/
```

PR #205, opened 2020-07-22
Remove unused public secrets from travis yml

bjrne:
In order to not confuse new users of whether this is a security vulnerability or whether to use them and actually accidentally expose their secrets.


maltezacharias (approved):
Been bugging me as well

Issue #33, opened 2020-06-17
Design welcoming README page


bjrne:
Design a README page that invites you to check out our project. Possible attributes are:
- Visually appealing
- Conveying a well-maintained state
- Clear steps for getting started
- Listing starting points for further reading instead of putting everything in the readme
- Using our colorscheme (tbd)
- Using our logo
- Showing off lots of project badges in the top ;)

This might include moving topics to a new github pages instance, to the wiki or into hierarchical readme's, but that might as well be another issue.

bjrne:
- **Introduction**
    - [ ] Netter Screenshot
    - [ ] Was ist es
    - [ ] Entstehung paar Sätze

- **Get Started**
    - [ ] Code-Beispiel aus den Docs und Link

- **Technical Documentation**
    - [ ] Link auf RTD

- **Contributing**
    - [ ] Link auf CONTRIBUTING.md

PR #206, opened 2020-07-22
[Documentation] Add docs for permissions and ENV vars

bjrne:
Are there any env variables missing? Maybe someone with more knowledge of priorities when choosing values from .env files and bash contexts can add a paragraph about that.


docs/configuration.rst
```diff
@@ -8,14 +8,25 @@ Environment variables
 -------------------------
 We use environment variables for:
 
-* Configuration
+* Configuration (mapview)
+* Credentials
 
-  * choosing a mapview backend that delivers the map tiles (square pngs of the map at a certain zoom level) to the user. For more technical information, refer to :ref:`architecture:mapview`.
+This is a complete list of environment variables we use. Their documentation is documented further in their respective topic:
```
bjrne, line 14: Is it though? Do we use more?

Baschdl: I cannot think of any other variables except the ones used in travis testing logic

Baschdl (approved):
Nicely written and it renders 👍

Baschdl:
@bjrne Did you leave the same sections on purpose in the README?

bjrne:
For now, yes, as we still work with it. A new readme is part of an extra issue, #33. And also some paragraphs still contain helpful information for writing other parts of the documentation, so I thought I'd leave it in. Replacing the whole Readme in the end is easy enough :)

PR #208, opened 2020-08-01
Enable favicon template in devMode

maltezacharias:
The favicon template introduced in #204  is only generated in production builds. Running the webpack build process in development mode will not create the needed files. This PR enables building the files in devMode as well.

The cleanUp Process (deletes dist/* on each build) is also modified not to delete the favicon files. This allows for the cache option to work and keeps the build process fast in devMode (except for the first one)

### DIFF
--- a/frontend/webpack.config.js
+++ b/frontend/webpack.config.js
@@ -141,7 +141,13 @@ module.exports = {
         })],
     },
     plugins: [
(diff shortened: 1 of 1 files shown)


feeds (approved):

bjrne:
Thanks, looks good!

PR #214, opened 2020-08-15
Good-looking edit profile page

feeds:


backend/apps/matching/configuration/participant.py
```diff
@@ -5,7 +5,22 @@
...
+                p.get_signup_layout(ignore_private=True) for p in self.properties
+            ]
+        return self.signup_layout_all
+
+    def view_layout(self):
+        if self.signup_layout_private is None:
+            self.signup_layout_private = [
+                p.get_signup_layout(ignore_private=False) for p in self.properties
+            ]
+        return self.signup_layout_private
```
bjrne, line 23: Maybe we can choose a better name? It is not clear what is private here, private like internal? Private like user-private? admin-private? authenticated-private?

bjrne (approved):
Names are refactorable :wink:
But nice overall.

backend/apps/matching/configuration/participant.py
bjrne: Well this was intended for this PR 🙈 or at least an explanation in a comment? I myself don't understand what it means 😅

Baschdl: @feeds

PR #215, opened 2020-08-16
Configurable Approvals

### LINKED ISSUE #186: Make validation requirements configurable
A future user should be able to setup easily, which users have to be approved and what consequences this has (eg what fields are visible with an without validation)

feeds:
- Make the approval configurable (= users are added to approved group on signup and fake user creation)
- Add a few new names instead of A here and there
- Fix the Signup

Partly fixes #186


backend/match4everyone/configuration/B.py
```diff
@@ -13,8 +13,8 @@ class BConfig(ParticipantConfig):
     #    NewPermissions.can_contact_type_a_if_approved,
     # ]
 
-    profile_visible_for_A = True
-    profile_visible_for_other_B = False
+    profile_visible_for_participants_of_different_type = True
+    profile_visible_for_participants_of_same_type = False
```
bjrne, line 18: Is this missing a
```suggestion
    needs_manual_approval_from_staff = True
```
line? :)

bjrne (approved):
Seems good!

backend/match4everyone/configuration/B.py
feeds: I set the default value to `True` in the superclass to make approvals the default behaviour, so it is not necessary.

bjrne: I think I would still put it here for a better understanding where to change it, don't you think?

PR #218, opened 2020-08-16
Fix name of env var

bjrne:
`TRAVIS_PULL_REQUEST_SLUG` is the `owner/repo_name` of the branch of whoever opened the pullrequest, so when an outsider of this organization opens a pull request, even though it runs in our travis, the code is not tested with sendgrid.

Moreover, that env var (PR_SLUG) only gets set for the PR-travis build, not for the branch travis build.

This new env var should be the solution.


Baschdl:
`TRAVIS_REPO_SLUG` is not set contradicting https://github.com/travis-ci/docs-travis-ci-com/blob/master/user/environment-variables.md

Baschdl:
Closed for now, seems to be a Travis bug preventing us from always testing with sendgrid

PR #219, opened 2020-08-21
Prevent required properties in conditionals

feeds:
We did not invest time to handle required properties in conditionals, which can make problems on signup. For now, we thus throw an error if someone sets up a configuration with this property.


bjrne (commented):
Untested, but code seems fine. Will test.

feeds:
@maltezacharias turns out the signup bug was already fixed here ;)

PR #223, opened 2020-09-03
fixes #222

### LINKED ISSUE #222: Create DB and delete script not working in docker
If running the ./scripts/delete_db_and_setup.sh scripts as shown in the docs it doesn't currently work in the docker environment. Docker will run in /backend while the script expects to run one directory above backend. The docker-compose command needs to be invoked with -w / to set the working dir to root, one level above /backend where the app is installed.

maltezacharias:
Adds parameter for working directory for the DB setup script invocation documentation


bjrne (approved):
Good, but it would be _so awesome_ to also see this added to the documentation :wink:

PR #225, opened 2020-09-03
224 browser auto fill

### LINKED ISSUE #224: UX: Make sure browser know which parts to auto-fill and password-save

maltezacharias:
Closes #224 

Introduces a new hidden (via css) field in the password reset form with the username and appropriate autocomplete settings so that credentials will be saved in the browser.

Considered if this was bad practice but considering that our password reset mail includes the e-mail as well as the password this should be good. Additionally with this changes users will be able to use autogenerated passwords as suggested by current browsers which also benefits account security

ToDo:
- [x] Check if same change needs to be applied to password change template


Baschdl:
Where is the password included in the mail?
https://github.com/match4everyone/match4everything/blob/3b5e9d856a644ca38b5ee3175bea1f2b9b8596ca/backend/apps/matching/templates/registration/password_reset_email_.html#L3-L14

Baschdl:
The new password on `password_change` is also not properly detected

maltezacharias:
> Where is the password included in the mail?

Sorry, I meant link to reset the password. 

> The new password on password_change is also not properly detected

I tested that and it updated fine here. (Tested on Windows&Chrome) with 64d4b5b8457d530ccafe0890383faf502a115d71
What did you use to test?

maltezacharias:
@Baschdl could you answer the above?

maltezacharias:
One more idea: This might not work on non-ssl pages, I tested with a local SSL Proxy and a properly signed certificate

Baschdl:
I used Firefox on Mac OS

Baschdl (approved):
Saving of username+password works, auto-generation of password doesn't work in Safari, Firefox and Chrome on Mac

PR #227, opened 2020-09-11
Set black to specific version

bjrne:
The new version ([20.8b1](https://github.com/psf/black/releases/tag/20.8b1)) does not respect our line length.


Baschdl (approved):
Beautiful

PR #228, opened 2020-10-04
Improve and document the configuration

feeds:
Small improvement of the configuration documentation (baby steps :))

Also, our app should not directly import "match4everything" for compatibility reasons, so we import the configuration from `django.conf` now


backend/apps/matching/models/user.py
```diff
@@ -46,15 +44,19 @@ def new(email, pwd=None, **kwargs):
 
         if kwargs["is_A"]:
             user.groups.add(group_is_a)
-            if B.profile_visible_for_participants_of_different_type:
+            if settings.PARTICIPANT_SETTINGS[
+                "B"
+            ].profile_visible_for_participants_of_different_type:
```
bjrne, line 49: Very ugly formatting :see_no_evil: 
I know its not your fault, it just looks awful.

docs/configuration.rst
```diff
@@ -27,9 +27,37 @@ This is a complete list of environment variables we use. Their documentation is
...
+
+where the latter are both an instance of your customized subclass of ``ParticipantConfig``.
+This class allows you to specify the information you would like to gather from each type of participant as well
+as the permissions of who can access which type of data.
+
+# Properties
+
+* ``name`` - the name with which this participant group should be referred to on the website itself
+
+* ``properties`` - the list of properties that the participant type should be able to set on signup. The available properties are documented below.
```
bjrne, line 49: A hyperlink could be useful in the finished PR :relaxed: 

```suggestion
* ``properties`` - the list of properties that the participant type should be able to set on signup. The available properties are documented below_.

.. _below: :ref:`configuration:properties`
```

bjrne (commented):
Seems good, will test locally when complete. I like the separation of the apps :)

